PR #10, opened 2022-08-16
Add hardware reboot cause as actual reboot cause when soft reboot fails

sujinmkang:
Why I did it
Add the hardware reboot cause when the previous software reboot failed

How I did it
Check both hardware reboot cause and software reboot cause.
Add the hardware reboot as actual reboot cause
if any hardware reboot cause is available for any software reboot.

How to verify it
Perform fast/warm/soft-reboot, cold reboot and hardware reboots and check the reboot-cause

### DIFF
--- a/scripts/determine-reboot-cause
+++ b/scripts/determine-reboot-cause
@@ -158,25 +158,7 @@ def get_reboot_cause_dict(previous_reboot_cause, comment, gen_time):
  
     return reboot_cause_dict
 
-
-def main():
-    # Configure logger to log all messages INFO level and higher
-    sonic_logger.set_min_log_priority_info()
-
-    sonic_logger.log_info("Starting up...")
-
-    if not os.geteuid() == 0:
-        sonic_logger.log_error("User {} does not have permission to execute".format(pwd.getpwuid(os.getuid()).pw_name))
-        sys.exit("This utility must be run as root")
-
-    # Create REBOOT_CAUSE_DIR if it doesn't exist
-    if not os.path.exists(REBOOT_CAUSE_DIR):
-        os.makedirs(REBOOT_CAUSE_DIR)
-
-    # Remove stale PREVIOUS_REBOOT_CAUSE_FILE if it exists
-    if os.path.exists(PREVIOUS_REBOOT_CAUSE_FILE):
-        os.remove(PREVIOUS_REBOOT_CAUSE_FILE)
-
+def determin_reboot_cause():
     # This variable is kept for future-use purpose. When proc_cmd_line/vendor/software provides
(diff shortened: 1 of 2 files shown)


prgeor:
@sujinmkang can you test this on Dell Z9332F
sync; sleep 2; sync; echo -c /proc/sysrq-trigger

See if you see the correct reboot cause

scripts/determine-reboot-cause
```diff
@@ -122,7 +122,12 @@ def find_hardware_reboot_cause():
     else:
         sonic_logger.log_info("No reboot cause found from platform api")
 
-    hardware_reboot_cause = "{} ({})".format(hardware_reboot_cause_major, hardware_reboot_cause_minor)
+    hardware_reboot_cause_minor_str = ""
+    if hardware_reboot_cause_minor:
+        hardware_reboot_cause_minor_str = " ({})".format(hardware_reboot_cause_minor)
+
+    hardware_reboot_cause = hardware_reboot_cause_major + hardware_reboot_cause_minor_str
```
prgeor, line 129: can we add space between the two strings ?

scripts/determine-reboot-cause
```diff
@@ -158,6 +163,50 @@ def get_reboot_cause_dict(previous_reboot_cause, comment, gen_time):
...
+        previous_reboot_cause = hardware_reboot_cause
+        # Check if any software reboot was issued before this hardware reboot happened
+        if software_reboot_cause is not REBOOT_CAUSE_UNKNOWN:
+            additional_reboot_info = software_reboot_cause
+        elif proc_cmdline_reboot_cause is not None:
+            additional_reboot_info = proc_cmdline_reboot_cause
+    elif proc_cmdline_reboot_cause is not None:
+        if software_reboot_cause is not REBOOT_CAUSE_UNKNOWN:
+            # Get the reboot cause from REBOOT_CAUSE_FILE
+            previous_reboot_cause = software_reboot_cause
```
prgeor, line 202: ignoring the proc_cmdline_reboot_cause which is what being checked at line 199?

scripts/determine-reboot-cause
sujinmkang: It's added when hardware_reboot_cause_minor_str was constructed.

scripts/determine-reboot-cause
sujinmkang: As I explained, this is for the fast/warm-reboot case. the software_reboot_cause contains the user and time information.

prgeor (approved):

prgeor:
@sujinmkang please make sure this is ported to 202205

PR #11, opened 2022-08-17
Enable TACACS accounting when enable TACACS authentication.

liuh-80:
#### Why I did it
To enable TACACS accounting when enable TACACS authentication.

#### How I did it
Listen TACACS authentication enable event and enable TACACS accounting when receive event.

#### How to verify it
Manually test.
Pass all E2E test case.

Enable TACACS accounting when enable TACACS authentication.

#### A picture of a cute animal (not mandatory but encouraged)


liuh-80:
Deprecate this PR, will enable accounting as default by config.

PR #13, opened 2022-08-19
[caclmgrd][chassis]: Fix missing acl rules to allow internal docker traffic from fabric namespaces

SuvarnaMeenakshi:
Signed-off-by: Suvarna Meenakshi <sumeenak@microsoft.com>
Requires https://github.com/sonic-net/sonic-buildimage/pull/11793
test_cacl_application fails on VoQ chassis Supervisor with the error:
Failed: Missing expected iptables rules: set(['-A INPUT -s 240.127.1.1/32 -d 240.127.1.1/32 -j ACCEPT', '-A INPUT -s 240.127.1.3/32 -d 240.127.1.1/32 -j ACCEPT', '-A INPUT -s 240.127.1.2/32 -d 240.127.1.1/32 -j ACCEPT'])
This failure is seen because acl rules to allow traffic from fabric namespaces is missing.
This PR is to include fabric namespace docker mgmt ips so that acl rules to allow traffic from namespace is added for fabric namespace as well.
Initial PR which added support for multi-asic for reference: https://github.com/sonic-net/sonic-buildimage/pull/5022

### DIFF
--- a/scripts/caclmgrd
+++ b/scripts/caclmgrd
@@ -141,15 +141,15 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
             self.update_thread[back_asic_namespace] = None
             self.lock[back_asic_namespace] = threading.Lock()
             self.num_changes[back_asic_namespace] = 0
-            self.update_docket_mgmt_ip_acl(back_asic_namespace)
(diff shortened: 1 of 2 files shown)


arlakshm:
Can you rebase your branch ?

judyjoseph (approved):
LGTM

PR #14, opened 2022-08-19
Added Support to render Feature Table using Device running metadata.

abdosi:
What I did:
Added Support to render Feature Table using Device running metadata
Also added support to render 'has_asic_scope' field of Feature Table.

Why I did:
Details: https://github.com/sonic-net/sonic-buildimage/pull/11796

How I verify:
Manual Verification
UT added including multi-asic,

### DIFF
--- a/scripts/hostcfgd
+++ b/scripts/hostcfgd
@@ -232,7 +232,7 @@ class FeatureHandler(object):
         """
         Summary:
         Updates the state field in the FEATURE|* tables as the state field
-        might have to be rendered based on DEVICE_METADATA table and generated Device Running Metadata
+        might have to be rendere dbased on DEVICE_METADATA table and generated Device Running Metadata
         """
         for feature_name in feature_table.keys():
             if not feature_name:
@@ -285,8 +285,7 @@ class FeatureHandler(object):
             self.disable_feature(feature)
             syslog.syslog(syslog.LOG_INFO, "Feature {} is stopped and disabled".format(feature.name))
 
-        if self.is_multi_npu:
-            self.sync_feature_asic_scope(feature)
+        self.sync_feature_asic_scope(feature)
 
         return True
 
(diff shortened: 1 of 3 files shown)


abdosi:
cc @arlakshm @judyjoseph

abdosi:
cc @anamehra @SuvarnaMeenakshi

abdosi:
@arlakshm can you please help review this.

arlakshm:
@abdosi , can you update the branch ?

scripts/hostcfgd
```diff
@@ -284,8 +294,47 @@ class FeatureHandler(object):
...
+                for cmd in cmds:
+                    syslog.syslog(syslog.LOG_INFO, "Running cmd: '{}'".format(cmd))
+                    try:
+                        run_cmd(cmd, raise_exception=True)
+                    except Exception as err:
+                        syslog.syslog(syslog.LOG_ERR, "Feature '{}.{}' failed to be stopped and disabled"
+                                      .format(feature.name, feature_suffixes[-1]))
+                        self.set_feature_state(feature, self.FEATURE_STATE_FAILED)
+                        return
+        self._config_db.mod_entry('FEATURE', feature_config.name, {'has_per_asic_scope': str(feature_config.has_per_asic_scope)})
```
arlakshm, line 336: The host config_db will have `has_per_asic_scope` as failed but the config_db for the asic will have them as enabled? should we update that as well to avoid confusion

mlok-nokia:
@arlakshm & @abdosi  Hi, Is this PR going be merged soon?   We need this fix to check what else we need to do to address some of  Error messages on Supervisor card for some of the feature such lldp0,lldp1...

scripts/hostcfgd
abdosi: @arlakshm yes, that we can do later if needed. Currently also other fields of this table are not same between host and asic.
I think that change will be coming as part of separate PR may be where we needed to update (ex:macsec feature state) that need be updated in asic db also. In that PR we can update this field also.

@judyjoseph

abdosi:
> @abdosi , can you update the branch ?

@arlakshm done

arlakshm (approved):

PR #21, opened 2022-09-26
[multi-asic] Fix the issue where if dest port is not specified in ACL rule than caclmgrd gets exception

abdosi:
What I did:
Fixes: https://github.com/sonic-net/sonic-buildimage/issues/12179

Why I did:
Fix the issue where if dest port is not specified in ACL rule than for multi-asic where we create NAT rule to forward traffic from Namespace to host fail with exception.

How I verify:-
Added UT for the scenario.


ZhaohuiS:
@abdosi Thank you for the quick fix. The change of EXTERNAL_CLIENT was also merged into 202012 and 202205 branch, could you please submit your change into these 2 branches too? Thanks a lot!

ZhaohuiS (approved):

anamehra:
Hi @abdosi  , need this fix on the 202205 branch as well. Thanks!

PR #23, opened 2022-10-26
Fix the exception issue when services is not specified in CTRLPLANE ACL

pettershao-ragilenetworks:
LOG:
root@sonic:/home/admin# config acl add table test CTRLPLANE
root@sonic:/home/admin# 

image


pettershao-ragilenetworks:
@qiluo-msft @prsunny please review， thanks

PR #24, opened 2022-11-02
Remove subprocess with shell=True

maipbui:
Signed-off-by: maipbui <maibui@microsoft.com>
#### Why I did it
`subprocess` is used with `shell=True`, which is very dangerous for shell injection.
#### How I did it
remove `shell=True`, use `shell=False`
#### How to verify it
Pass UT
Manual test


scripts/hostcfgd
```diff
@@ -86,7 +88,17 @@ def signal_handler(sig, frame):
...
-        subprocess.check_call(cmd, shell=True)
+        subprocess.check_call(cmd)
+    except Exception as err:
+        if log_err:
+            syslog.syslog(syslog.LOG_ERR, "{} - failed: return code - {}, output:\n{}"
+                  .format(err.cmd, err.returncode, err.output))
+        if raise_exception:
+            raise
+
+def run_cmd_pipe(cmd0, cmd1, cmd2, log_err=True, raise_exception=False):
```
qiluo-msft, line 99: >run_cmd_pipe [](http://example.com/codeflow?start=4&length=12)

Similar function already implemented. Could you reuse in the scope of this repo?

scripts/hostcfgd
```diff
@@ -759,8 +771,12 @@ class AaaCfg(object):
 
     def modify_single_file(self, filename, operations=None):
         if operations:
-            cmd = "sed -e {0} {1} > {1}.new; mv -f {1} {1}.old; mv -f {1}.new {1}".format(' -e '.join(operations), filename)
-            os.system(cmd)
+            e_operations_str = ' -e '.join(operations)
+            e_operations_list = split(e_operations_str)
```
qiluo-msft, line 775: >e_operations_list [](http://example.com/codeflow?start=12&length=17)

operations is already a list, no need to join and then split.

scripts/hostcfgd
```diff
@@ -759,8 +771,12 @@ class AaaCfg(object):
...
+            with open(filename+'new', 'w') as f:
```
qiluo-msft, line 776: >new [](http://example.com/codeflow?start=32&length=3)

you miss `.` ?

scripts/hostcfgd
```diff
@@ -971,9 +987,11 @@ class PasswHardening(object):
 
     def modify_single_file_inplace(self, filename, operations=None):
         if operations:
-            cmd = "sed -i {0} {1}".format(' -i '.join(operations), filename)
+            i_operations_str = ' -i '.join(operations)
+            i_operations_list = split(i_operations_str)
```
qiluo-msft, line 991: >split [](http://example.com/codeflow?start=32&length=5)

The same.

scripts/procdockerstatsd
```diff
@@ -135,7 +136,13 @@ class ProcDockerStats(daemon_base.DaemonBase):
         return True
 
     def update_processstats_command(self):
-        data = self.run_command("ps -eo uid,pid,ppid,%mem,%cpu,stime,tty,time,cmd --sort -%cpu | head -1024")
+        cmd = "ps -eo uid,pid,ppid,%mem,%cpu,stime,tty,time,cmd --sort -%cpu | head -1024"
+        cmd0 = ["ps", "-eo uid,pid,ppid,%mem,%cpu,stime,tty,time,cmd", "--sort", "-%cpu"]
+        cmd1 = ["head", "-1024"]
+        exitcode, data = getstatusoutput_noshell_pipe(cmd0, cmd1)
+        if exitcode != [0, 0]:
```
qiluo-msft, line 143: >exitcode [](http://example.com/codeflow?start=11&length=8)

If you want to check all array elements are zero, you can use `not exitcode.any()`

scripts/hostcfgd
maipbui: I think the list is joined by `' -e '`, not whitespace?

scripts/hostcfgd
maipbui: which function are you referring to?

tests/caclmgrd/caclmgrd_bfd_test.py
```diff
@@ -44,6 +44,8 @@ def test_caclmgrd_bfd(self, test_name, test_data, fs):
             call_rc = test_data["call_rc"]
             mocked_subprocess.call.return_value = call_rc
 
+            self.caclmgrd.ControlPlaneAclManager.get_namespace_mgmt_ip = mock.MagicMock()
+            self.caclmgrd.ControlPlaneAclManager.get_namespace_mgmt_ipv6 = mock.MagicMock()
```
ganglyu, line 48: Why do you add mock? And why can original unit test work without mock?

scripts/caclmgrd
```diff
@@ -178,7 +182,7 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
             commands: List of strings, each string is a shell command
```
ganglyu, line 182: It should be List of List?

scripts/caclmgrd
```diff
@@ -188,6 +192,19 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
                 return stdout.rstrip('\n')
         return ""
 
+    def run_commands_pipe(self, cmd0, cmd1, cmd2, cmd3):
+        """
+        Given a list of shell commands, run them in order
+        Args:
+            commands: List of strings, each string is a shell command
```
ganglyu, line 199: Still List of strings?

tests/caclmgrd/caclmgrd_bfd_test.py
maipbui: It throws this error: 
```
scripts/caclmgrd:202: in run_commands_pipe
    exitcodes, stdout = getstatusoutput_noshell_pipe(cmd0, cmd1, cmd2, cmd3)
/usr/local/lib/python3.9/dist-packages/sonic_py_common/general.py:52: in getstatusoutput_noshell_pipe
    popens = [Popen(cmd0, stdout=PIPE, universal_newlines=True)]
[OSError]: [Error 9] Bad file descriptor in the fake filesystem: '11'
```
I think it is due to 

> If close_fds is true, all file descriptors except 0, 1 and 2 will be closed before the child process is executed. Otherwise when close_fds is false, file descriptors obey their inheritable flag as described in [Inheritance of File Descriptors](https://docs.python.org/3/library/os.html#fd-inheritance).

in https://docs.python.org/3/library/subprocess.html#popen-constructor
Not sure how original UT works w/o mock

scripts/hostcfgd
qiluo-msft: I got it. Better to implement the transformation on a list directly.

scripts/hostcfgd
qiluo-msft: You implemented twice `run_commands_pipe` in this PR. I understand they are different a little bit. Could you try unify and reuse?

fastiuk, line 99: Why use that code and not use the python 'sh' library?
It supports everything you need here: shell lexicography and pipe.
It is just much easier to use it.
[Link to docs](https://amoffat.github.io/sh/)

tests/caclmgrd/caclmgrd_bfd_test.py
ganglyu: You have changed the behavior of this unit test, and we might have better solution.
https://github.com/sonic-net/sonic-host-services/blob/master/tests/caclmgrd/caclmgrd_bfd_test.py#L37
Above line doesn't work with your commit, and I think you should fix this mock and keep the original behavior.

maipbui: I fixed the mock and revert to original behavior. Could you help review again this UT?

ganglyu (approved):

tests/caclmgrd/caclmgrd_dhcp_test.py
```diff
@@ -44,6 +44,8 @@ def test_caclmgrd_dhcp(self, test_name, test_data, fs):
 
             mark = test_data["mark"]
 
+            self.caclmgrd.ControlPlaneAclManager.get_namespace_mgmt_ip = mock.MagicMock()
+            self.caclmgrd.ControlPlaneAclManager.get_namespace_mgmt_ipv6 = mock.MagicMock()
```
ganglyu, line 48: same issue

tests/caclmgrd/caclmgrd_feature_test.py
```diff
@@ -44,6 +44,8 @@ def test_feature_present(self, test_name, test_data, fs):
             call_rc = test_data["call_rc"]
             mocked_subprocess.call.return_value = call_rc
 
+            self.caclmgrd.ControlPlaneAclManager.get_namespace_mgmt_ip = mock.MagicMock()
+            self.caclmgrd.ControlPlaneAclManager.get_namespace_mgmt_ipv6 = mock.MagicMock()
```
ganglyu, line 48: same issue

tests/caclmgrd/caclmgrd_bfd_test.py
ganglyu: Other file is still using mock for get_namespace_mgmt_ip and get_namespace_mgmt_ipv6

maipbui: Fixed. And thanks so much for your suggestion!

ganglyu (approved):

scripts/hostcfgd
maipbui: We want to have a more secure codebase and aim to fix potentially dangerous function calls that could introduce security vulnerabilities if used incorrectly. I fixed the functions based on Semgrep recommendations (a static analysis tool to find security issues, some refs: https://semgrep.dev/docs/cheat-sheets/python-command-injection/, https://github.com/sonic-net/sonic-buildimage/actions/runs/3412786426/jobs/5678731076). Not sure if we want to use third-party library and the 'sh' library docs does not mention anything about security.

maipbui: Unified duplicate code to one function.

scripts/caclmgrd
```diff
@@ -154,40 +154,58 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
...
-            proc = subprocess.Popen(cmd, shell=True, universal_newlines=True, stdout=subprocess.PIPE)
+            proc = subprocess.Popen(cmd, universal_newlines=True, stdout=subprocess.PIPE)
 
             (stdout, stderr) = proc.communicate()
-
-            if proc.returncode != 0:
-                self.log_error("Error running command '{}'".format(cmd))
-            elif stdout:
-                return stdout.rstrip('\n')
+            self.log_output(cmd, [proc.returncode], stdout)
```
qiluo-msft, line 194: >log_output [](http://example.com/codeflow?start=17&length=10)

This extracted function `log_output` change the original behavior: `return` will terminate for loop immediately. you new code will always return empty string in the outside function.

qiluo-msft: You are allowed not to extract a function in this PR, I only see 2 use cases.

qiluo-msft:
>         # Initlaize Global config that loads all database*.json

Seem like not related to the PR title. Do you really need it? If yes, how does this new requirement introduced?

---
Refers to: scripts/hostcfgd:217 in 592079c. [](commit_id = 592079c5753e610dabec68bcf85c1d1ffee11bf3, deletion_comment = False)

scripts/caclmgrd
```diff
@@ -154,38 +154,59 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
...
-        ipv6_address_get_command = iptable_ns_cmd_prefix + "ip -6 -o addr show scope global " + ("eth0" if namespace else "docker0") +\
-                                 " | awk '{print $4}' | cut -d'/' -f1 | head -1"
-        return self.run_commands([ipv6_address_get_command])
+        ipv6_address_cmd0 = iptable_ns_cmd_prefix + ['ip', '-6', '-o', 'addr', 'show', 'scope', 'global', ("eth0" if namespace else "docker0")]
+        ipv6_address_cmd1 = ['awk', '{print $4}']
+        ipv6_address_cmd2 = ['cut', '-d', '/', '-f1']
+        ipv6_address_cmd3 = ['head', '-1']
+        return self.run_commands_pipe(ipv6_address_cmd0, ipv6_address_cmd1, ipv6_address_cmd2, ipv6_address_cmd3)
+
+    def log_output(self, cmd, exitcodes, stdout):
```
qiluo-msft, line 178: >log_output [](http://example.com/codeflow?start=8&length=10)

Add an explicit `return None` at the function end, since you are expecting the behavior.

scripts/procdockerstatsd
```diff
@@ -135,7 +136,13 @@ class ProcDockerStats(daemon_base.DaemonBase):
         return True
 
     def update_processstats_command(self):
-        data = self.run_command("ps -eo uid,pid,ppid,%mem,%cpu,stime,tty,time,cmd --sort -%cpu | head -1024")
+        cmd = "ps -eo uid,pid,ppid,%mem,%cpu,stime,tty,time,cmd --sort -%cpu | head -1024"
```
qiluo-msft, line 139: >cmd = "ps -eo uid,pid,ppid,%mem,%cpu,stime,tty,time,cmd --sort -%cpu | head -1024" [](http://example.com/codeflow?start=8&length=82)

Assume someone fix cmd0, but forget to change cmd, the log_error message will be wrong.

scripts/hostcfgd
```diff
@@ -550,12 +561,12 @@ class Iptables(object):
...
                 do nothing if rule exists
                 '''
-                ret = subprocess.call(cmd, shell=True)
+                ret = subprocess.call(cmd)
                 if ret == 0:
                     syslog.syslog(syslog.LOG_INFO, "{} rule exists in {}".format(ip, chain))
                 else:
                     # Modify command from Check to Append
-                    iptables_cmds.append(cmd.replace("check", "append"))
+                    iptables_cmds.append([word if word != "--check" else "--append" for word in cmd])
```
qiluo-msft, line 569: >cmd [](http://example.com/codeflow?start=96&length=3)

Strictly speaking, you changed behavior.
Suggest loop through cmd, replace in each element.

qiluo-msft (approved):

tests/mock_connector.py
```diff
@@ -22,3 +22,10 @@ def keys(self, db_id, pattern):
 
     def get_all(self, db_id, key):
         return MockConnector.data[key]
+
+    def delete_all_by_pattern(self, db_id, pattern):
```
qiluo-msft, line 26: >delete_all_by_pattern [](http://example.com/codeflow?start=8&length=21)

This mock function only work on special use case, not a general implementation. Please consider copy the logic from codebase, and maybe implement a mock delete() function.

StormLiangMS:
@dgsudharsan  included the #34, which is newer version, I think we don't need this PR.

PR #25, opened 2022-11-28
Avoid printing message in error level when DEVICE_METADATA|localhost updates

stephenxs:
There is error message logged on receiving `DEVICE_METADATA|localhost.hostname` update with the same value `DeviceMetaCfg.hostname`.
However, this can happen when other fields in `DEVICE_METADATA|localhost` are updated after `hostname` is recorded by `DeviceMetaCfg` because it always notifies of updates of all the fields together.
In this case, `INFO` is preferred.

Signed-off-by: Stephen Sun <stephens@nvidia.com>

### DIFF
--- a/scripts/hostcfgd
+++ b/scripts/hostcfgd
@@ -1355,7 +1355,13 @@ class DeviceMetaCfg(object):
 
         # Restart hostname-config service when hostname was changed.
         # Empty not allowed
-        if new_hostname and new_hostname != self.hostname:
+        if not new_hostname:
+            syslog.syslog(syslog.LOG_ERR,
+                          'Hostname was not updated: Empty not allowed')
+        elif new_hostname == self.hostname:
+            syslog.syslog(syslog.LOG_INFO,
+                          'Hostname was not updated: Already set up with the same name: {}'.format(self.hostname))
+        else:
             syslog.syslog(syslog.LOG_INFO, 'DeviceMetaCfg: Set new hostname: {}'
                                            .format(new_hostname))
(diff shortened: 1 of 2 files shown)


scripts/hostcfgd
```diff
@@ -1369,8 +1369,12 @@ class DeviceMetaCfg(object):
             run_cmd('sudo monit reload')
         else:
             msg = 'Hostname was not updated: '
-            msg += 'Already set up' if new_hostname else 'Empty not allowed'
-            syslog.syslog(syslog.LOG_ERR, msg)
+            if new_hostname:
+                msg += 'Already set up'
```
saiarcot895, line 1373: Could this maybe be modified to indicate that it's the same hostname?

stephenxs: Updated. Can you please review it again? Thanks.

saiarcot895 (approved):

scripts/hostcfgd
```diff
@@ -1369,8 +1369,12 @@ class DeviceMetaCfg(object):
             run_cmd('sudo monit reload')
         else:
```
qiluo-msft, line 1370: >else [](http://example.com/codeflow?start=8&length=4)

Suggest to flatten the branches as
```
if not new_hostname:
    ...
elif new_hostname == self.hostname:
    ...
else:
    ...
```

stephenxs: Updated. Thanks

ganglyu (approved):

qiluo-msft (approved):

PR #30, opened 2023-01-06
Replace /dev/stdin

ganglyu:
# Why I did it

When we use /dev/stdin as input file, CLI command can only read file for one time.
https://github.com/sonic-net/sonic-utilities/pull/2529
This PR will validate config file at first, and then CLI command will read config file for more than one time, so /dev/stdin does not work with latest code.

# How I did it

Use temp file to replace /dev/stdin

# How to verify it

Run unit test.

### DIFF
--- a/host_modules/config_engine.py
+++ b/host_modules/config_engine.py
@@ -2,6 +2,7 @@
 
 from host_modules import host_service
 import subprocess
+import os
 
 MOD_NAME = 'config'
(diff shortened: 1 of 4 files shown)


qiluo-msft:
Let's hold on this PR and revisit the behavior trigger PR https://github.com/sonic-net/sonic-utilities/pull/2529

PR #34, opened 2023-01-13
Remove subprocess with shell=True

maipbui:
Signed-off-by: maipbui <maibui@microsoft.com>
#### Why I did it
`subprocess` is used with `shell=True`, which is very dangerous for shell injection.
#### How I did it
remove `shell=True`, use `shell=False`
#### How to verify it
Pass UT
Manual test


scripts/procdockerstatsd
```diff
@@ -135,7 +136,13 @@ class ProcDockerStats(daemon_base.DaemonBase):
         return True
 
     def update_processstats_command(self):
-        data = self.run_command("ps -eo uid,pid,ppid,%mem,%cpu,stime,tty,time,cmd --sort -%cpu | head -1024")
+        cmd0 = ["ps", "-eo", "uid,pid,ppid,%mem,%cpu,stime,tty,time,cmd", "--sort", "-%cpu"]
+        cmd1 = ["head", "-1024"]
+        cmd = ' | '.join([' '.join(cmd0), ' '.join(cmd1)])
```
qiluo-msft, line 141: >cmd [](http://example.com/codeflow?start=8&length=3)

You can move `cmd` to inside `if` block, so it will optimize performance a little bit.

tests/mock_connector.py
```diff
@@ -22,3 +22,13 @@ def keys(self, db_id, pattern):
 
     def get_all(self, db_id, key):
         return MockConnector.data[key]
+
+    def delete(self, db_id, key):
+        return MockConnector.data.delete(db_id, key)
+
+    def delete_all_by_pattern(self, db_id, pattern):
+        client = MockConnector.data
+        keys = client.keys()
```
qiluo-msft, line 31: >keys [](http://example.com/codeflow?start=22&length=4)

You should use `pattern`.

tests/mock_connector.py
```diff
@@ -22,3 +22,13 @@ def keys(self, db_id, pattern):
 
     def get_all(self, db_id, key):
         return MockConnector.data[key]
+
+    def delete(self, db_id, key):
+        return MockConnector.data.delete(db_id, key)
```
qiluo-msft, line 27: >db_id [](http://example.com/codeflow?start=41&length=5)

Seem db_id is not used with `data`

qiluo-msft (approved):

qiluo-msft (approved):

PR #35, opened 2023-01-13
[sonic-host-server] Simplify `register_dbus`

maipbui:
The goal is to prevent use `importlib.import_module` function.
Verified by manual test.


qiluo-msft (approved):

ganglyu:
Why can't we use import_module?

ganglyu:
Can we use __import__ instead?

ganglyu (approved):

PR #36, opened 2023-01-16
Check config file not empty after modify it in hostcfgd.

liuh-80:
**What I did**
Check /etc/pam.d/sshd integrity after modify it in hostcfgd.

**Why I did it**
Found some incident that /etc/pam.d/sshd become empty file during OR upgrade. 

**How I verified it**
Pass all UT.
Add new UT to cover new code.

**Details if related**


scripts/hostcfgd
```diff
@@ -781,11 +792,29 @@ class AaaCfg(object):
                 interface_ip = ipv4_addr
         return interface_ip
 
+
+    def check_file_integrity(self, filename):
```
qiluo-msft, line 796: >check_file_integrity [](http://example.com/codeflow?start=8&length=20)

Strictly speaking, this file is only check file non-empty. If this is expected and good enough, please change the function name to narrow scope.

Otherwise, you may find some command line to verify the config. For example, `sshd -T` could be used to verify `/etc/ssh/sshd_config`. I am not sure if you can find the verification command for any other command.

scripts/hostcfgd
```diff
@@ -781,11 +792,29 @@ class AaaCfg(object):
...
+        rc, out = exec_cmd(cmd="du -b {}".format(filename), verbose=False)
```
qiluo-msft, line 797: >exec_cmd [](http://example.com/codeflow?start=18&length=8)

This is python. So prefer python API than equivalent command line invoke

scripts/hostcfgd
liuh-80: Fixed by change function name.

scripts/hostcfgd
liuh-80: Fixed.

scripts/hostcfgd
```diff
@@ -781,11 +781,29 @@ class AaaCfg(object):
...
         return interface_ip
 
+
+    def check_file_not_empty(self, filename):
+        exists = os.path.exists(filename)
+        if not exists:
+            syslog.syslog(syslog.LOG_ERR, "file size check failed: {} is missing".format(filename))
+            return
+
+        # du command result is: 'size\tpath'
```
qiluo-msft, line 791: ># du command result is: 'size\tpath' [](http://example.com/codeflow?start=8&length=36)

The comment is not relevant any more. Please remove it.

liuh-80: Fixed.

qiluo-msft (approved):

liuh-80:
Manually cherry-pick PR created: https://github.com/sonic-net/sonic-buildimage/pull/14115

PR #37, opened 2023-01-18
Change file permission

maipbui:
Restrict file permission access.

### DIFF
--- a/scripts/aaastatsd
+++ b/scripts/aaastatsd
@@ -151,7 +151,7 @@ class RadiusStatistics:
                 os.unlink(stats_file)
         else:
             open(stats_file, 'a').close()
-            os.chmod(stats_file, 0o600)
+            os.chmod(stats_file, 0o644)
             self.handle_update(addr)
 
     def handle_clear(self):
--- a/scripts/aaastatsd
+++ b/scripts/aaastatsd
@@ -151,7 +151,7 @@ class RadiusStatistics:
                 os.unlink(stats_file)
         else:
             open(stats_file, 'a').close()
-            os.chmod(stats_file, 0o666)
+            os.chmod(stats_file, 0o600)
             self.handle_update(addr)
 
     def handle_clear(self):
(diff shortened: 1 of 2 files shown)


maipbui:
@a-barboza Could you help review this PR?

a-barboza:
> @a-barboza Could you help review this PR?

@maipbui Do you want to check if 0600 will cause issue if read permission is required by non-root users? Sometimes, processes without privileges need to be able to read these files, I think. In that case 0644 might be better. In any case I do not have any objections to the stats file having 0600 permissions.

maipbui:
> > @a-barboza Could you help review this PR?
> 
> @maipbui Do you want to check if 0600 will cause issue if read permission is required by non-root users? Sometimes, processes without privileges need to be able to read these files (PAM configuration files), I think. In that case 0644 might be better. In any case I do not have any objections to the stats file having 0600 permissions.

Yes, I do want to check and verify the change. Could you help verify?

scripts/hostcfgd
```diff
@@ -1016,7 +1016,7 @@ class PasswHardening(object):
         # Use rename(), which is atomic (on the same fs) to avoid empty file
         with open(PAM_PASSWORD_CONF + ".tmp", 'w') as f:
             f.write(pam_passwh_conf)
-        os.chmod(PAM_PASSWORD_CONF + ".tmp", 0o644)
+        os.chmod(PAM_PASSWORD_CONF + ".tmp", 0o600)
```
qiluo-msft, line 1019: >0o600 [](http://example.com/codeflow?start=45&length=5)

I checked the permission for `/etc/pam.d/common-password` is 644. So suggest chmod to the same permission on .tmp file.

scripts/aaastatsd
```diff
@@ -151,7 +151,7 @@ class RadiusStatistics:
                 os.unlink(stats_file)
         else:
             open(stats_file, 'a').close()
-            os.chmod(stats_file, 0o666)
+            os.chmod(stats_file, 0o600)
```
qiluo-msft, line 154: >0o600 [](http://example.com/codeflow?start=33&length=5)

We need to check original file permission.

qiluo-msft (approved):

PR #42, opened 2023-02-15
Added IP Table rule to allow eth1-midplane traffic for chassis

abdosi:
What I did:
Added IP Table rule to make sure we do not drop chassis internal traffic on eth1-midpplane when Control Plane ACL's are installed.

Why I did:
When Control Plane ACL's are installed there is  default Catch All rule is added to drop all traffic that is not white-listed explicitly https://github.com/sonic-net/sonic-host-services/blob/master/scripts/caclmgrd#L735. In this case Internal Traffic between Supervisor and LC will get drop. To fix this added explicit rule to allow all traffic coming from `eth1-midplane`.

Also added rule to allow traffic with source and destination as eth1-midplane ip which is used in redis communication on chassis supervisor.

How I verify:
Manually Verified rule is getting installed
sonic-mgmt case will be updated to account for this new iptable rule.

### DIFF
--- a/scripts/caclmgrd
+++ b/scripts/caclmgrd
@@ -296,17 +296,23 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
 
         return block_ip2me_cmds
 
-    def check_chassis_midplane_interface_exist(self):
-        ip_link_cmd0 = ['ip', 'link', 'show']
-        ip_link_cmd1 = ['grep', '-w', "eth1-midplane"]
-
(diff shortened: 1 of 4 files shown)


arlakshm (approved):

judyjoseph (approved):

abdosi:
@xumia why the PR checker stuck here ?

xumia:
> @xumia why the PR checker stuck here ?

The azure pipeline has an issue, @liushilongbuaa will fix it.

abdosi:
@xumia any update on this

PR #43, opened 2023-02-23
Check SONiC dependencies before installation.

liushilongbuaa:
Why I did it
SONiC related packages shouldn't be intalled from Pypi.
It is security compliance requirement.

How I did it
Check SONiC related packages when using setup.py.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,23 @@
+from __future__ import print_function
+import sys
 from setuptools import setup
+import pkg_resources
+from packaging import version
+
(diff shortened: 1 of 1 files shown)


xumia (approved):

xumia:
LGTM, @qiluo-msft , do you think it is good?

qiluo-msft (approved):

qiluo-msft:
This file is only used in build time or unit test, not in runtime. We can safely override the coverage checker.

PR #45, opened 2023-02-23
[config reload]Config Reload Enhancement

dgsudharsan:
Code changes for config reload Enhancement https://github.com/sonic-net/SONiC/pull/1203.
Enhancing config reload to sequence the services and faster system initialization.

Immediately restart the critical services during config reload.
The non critical should be started only after all the ports are initialized.
Services can be configured to be started immediately or delayed. This can be using a field in FEATURE table.
The existing timers should be removed by this event driven approach.
This flow is applicable in case of all reboots (warm/fast/cold) as well as config reload.

### DIFF
--- a/scripts/hostcfgd
+++ b/scripts/hostcfgd
@@ -158,6 +158,9 @@ class Feature(object):
             feature_cfg (dict): Feature CONFIG_DB configuration
             deviec_config (dict): DEVICE_METADATA section of CONFIG_DB
         """
+        if 'has_timer' in feature_cfg:
+            syslog.syslog(syslog.LOG_ERR, "Invalid obsolete field 'has_timer' in FEATURE table. Please update configuration schema version")
+            raise Exception("Obsolte field 'has_timer' in FEATURE table")
 
         self.name = feature_name
         self.state = self._get_feature_table_key_render_value(feature_cfg.get('state'), device_config or {}, ['enabled', 'disabled', 'always_enabled', 'always_disabled'])
--- a/scripts/hostcfgd
+++ b/scripts/hostcfgd
@@ -206,42 +206,6 @@ class FeatureHandler(object):
     FEATURE_STATE_DISABLED = "disabled"
     FEATURE_STATE_FAILED = "failed"
 
-    def enableDelayedServices(self):
-        self.lock.acquire()
-        self.enable_delayed_service = True
-        for feature_name in self._cached_config:
-            if self._cached_config[feature_name].delayed:
-                self.update_feature_state(self._cached_config[feature_name])
-        self.lock.release()
-
-    def portListener(self):
-        syslog.syslog(syslog.LOG_INFO, "Starting port listener")
-        appl_db_connector = DBConnector("APPL_DB", 0)
-        appl_db_id = SonicDBConfig.getDbId("APPL_DB")
-        subscribe_port = swsscommon.SubscriberStateTable(appl_db_connector, "PORT_TABLE")
-        sel = swsscommon.Select()
-        sel.addSelectable(subscribe_port)
-
-        if (self.is_advanced_boot):
-             syslog.syslog(syslog.LOG_INFO, "Updating delayed features after warm/fast boot")
-             self.enableDelayedServices()
-
-        while True:
-            (state, selectableObj) = sel.select(PORT_INIT_TIMEOUT_SEC*1000)
-            # Continue if select is timeout or selectable object is not return
-            if state != swsscommon.Select.OBJECT:
-                if not self.enable_delayed_service:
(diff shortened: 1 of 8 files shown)


scripts/hostcfgd
```diff
@@ -160,7 +162,7 @@ class Feature(object):
         self.name = feature_name
         self.state = self._get_feature_table_key_render_value(feature_cfg.get('state'), device_config or {}, ['enabled', 'disabled', 'always_enabled', 'always_disabled'])
         self.auto_restart = feature_cfg.get('auto_restart', 'disabled')
-        self.has_timer = safe_eval(feature_cfg.get('has_timer', 'False'))
+        self.delayed = safe_eval(feature_cfg.get('delayed', 'False'))
```
qiluo-msft, line 165: >delayed [](http://example.com/codeflow?start=50&length=7)

The ConfigDB schema change it not backward compatible. Could you support both schema in this daemon? Or, could you check the schema version and fail immediately if old version?

scripts/hostcfgd
```diff
@@ -204,12 +206,18 @@ class FeatureHandler(object):
...
+    def __init__(self, config_db, feature_state_table, device_config, is_advanced_boot):
         self._config_db = config_db
         self._feature_state_table = feature_state_table
         self._device_config = device_config
         self._cached_config = {}
         self.is_multi_npu = device_info.is_multi_npu()
+        self.enable_delayed_service = False
+        self.is_advanced_boot = is_advanced_boot
+        self.lock = threading.Lock()
+        self.port_listener_thread = threading.Thread(target=self._portListener, name='port_listener_thread')
```
qiluo-msft, line 218: >Thread [](http://example.com/codeflow?start=46&length=6)

We discourage using threads. Could you explore if you can use `select` to implement? I understand rewrite the ConfigDB subscribe part need extra effort.

dgsudharsan: We previously used the subscribertable approach for which we were requested to move to configdbconnector to use a single FD. We cannot use configDBconnector for this scenario as the PORT_TABLE:PortInitdone is part of APPL_DB.

scripts/hostcfgd
dgsudharsan: Added the check as you requested

dgsudharsan:
@xumia @qiluo-msft Can you please restart the pipeline here?

xumia:
@dgsudharsan, the azp has an issue relative to the GitHub connection, fixed.

xumia:
https://dev.azure.com/mssonic/build/_build/results?buildId=232038&view=results

scripts/hostcfgd
```diff
@@ -156,11 +158,14 @@ class Feature(object):
             feature_cfg (dict): Feature CONFIG_DB configuration
             deviec_config (dict): DEVICE_METADATA section of CONFIG_DB
         """
+        if 'has_timer' in feature_cfg:
+            syslog.syslog(syslog.LOG_ERR, "Invalid obsolete field 'has_timer' in FEATURE table. Please update configuration schema version")
+            raise Exception("Obsolte field 'has_timer' in FEATURE table")
```
qiluo-msft, line 163: >Exception [](http://example.com/codeflow?start=18&length=9)

Suggest use specific type like ValueError.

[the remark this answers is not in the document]
qiluo-msft: Above syslog message is pretty good. Use it again in exception message, you may consider use a string var.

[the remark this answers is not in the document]
dgsudharsan: Addressed

dgsudharsan: Addressed

qiluo-msft (approved):

scripts/hostcfgd
```diff
@@ -156,11 +158,15 @@ class Feature(object):
             feature_cfg (dict): Feature CONFIG_DB configuration
             deviec_config (dict): DEVICE_METADATA section of CONFIG_DB
         """
+        if 'has_timer' in feature_cfg:
```
wenyiz2021, line 161: hi @dgsudharsan , wonder why you don't want 'has_timer' in config_db.json, FEATURE section?

dgsudharsan: @wenyiz2021 Please refer to this comment https://github.com/sonic-net/sonic-host-services/pull/45#discussion_r1119395605

dgsudharsan: Yes. If the config is part of upgrade, it will be automatically modified by the db_migrator. However if the user explicitly sets has_timer and performs a config reload in an image, it is expected to fail.

wenyiz2021: thanks for clarifying

PR #50, opened 2023-03-20
[caclmgrd]Allow Vxlan udp port on receiving vxlan tunnel configuration

dgsudharsan:
Currently VxLAN packets are dropped due to absence of rule in iptables. When there is L3 traffic on egress VTEP with no ARP entry, due to the absence of processing by kernel, the traffic gets dropped.
Instead the VxLAN packets should be processed and ARP request needs to be resolved by the kernel.

Made changes to allow VxLAN udp port when VxLAN tunnel is configured.

Note: This change doesn't consider the VxLAN_PORT configuration in the switch table. This use case is for the VNET and for EVPN the default value of 4789 is used.

### DIFF
--- a/scripts/caclmgrd
+++ b/scripts/caclmgrd
@@ -833,14 +833,18 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
         iptables_cmds = []
 
         # Add iptables/ip6tables commands to allow VxLAN packets
-        if ":" in self.VxlanSrcIP:
-            iptables_cmds.append(self.iptables_cmd_ns_prefix[namespace] + ['ip6tables', '-I', 'INPUT', '2', '-p', 'udp', '-d', self.VxlanSrcIP, '--dport', '4789', '-j', 'ACCEPT'])
-        else:
-            iptables_cmds.append(self.iptables_cmd_ns_prefix[namespace] + ['iptables', '-I', 'INPUT', '2', '-p', 'udp', '-d', self.VxlanSrcIP, '--dport', '4789', '-j', 'ACCEPT'])
+        ip_addr = ipaddress.ip_address(self.VxlanSrcIP)
+        if isinstance(ip_addr, ipaddress.IPv6Address):
+            iptables_cmds.append(self.iptables_cmd_ns_prefix[namespace] +
+                    ['ip6tables', '-I', 'INPUT', '2', '-p', 'udp', '-d', self.VxlanSrcIP, '--dport', '4789', '-j', 'ACCEPT'])
+        elif isinstance(ip_addr, ipaddress.IPv4Address):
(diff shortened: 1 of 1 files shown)


dgsudharsan:
@prsunny @srj102 Can you please review this change?

dgsudharsan:
The coverage is not met due to the absence of mock for generic swss library components which is missing today. I have added tests to cover the APIs introduced.

scripts/caclmgrd
```diff
@@ -115,6 +116,8 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
 
     DualToR = False
     bfdAllowed = False
+    VxlanAllowed = False
```
ZhaohuiS, line 119: @dgsudharsan  Is this VxlanAllowed variable used?

dgsudharsan: Yes. It is used in line number 992

scripts/caclmgrd
```diff
@@ -817,6 +820,46 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
...
+                break
+
+        if not self.VxlanSrcIP:
+            self.log_info("Received vxlan tunnel configuration without source ip")
+            return False
+
+        iptables_cmds = []
+
+        # Add iptables/ip6tables commands to allow VxLAN packets
+        if ":" in self.VxlanSrcIP:
```
ZhaohuiS, line 836: @dgsudharsan  Is it better to use this statement to check if it's ipv4 or ipv6? 
`if isinstance(ip_addr, ipaddress.IPv4Address):`

scripts/caclmgrd
```diff
@@ -817,6 +820,46 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
...
+
+    def block_vxlan_port(self, namespace):
+        if not self.VxlanSrcIP:
+            self.log_info("Cannot remove vxlan tunnel configuration without source ip")
+            return False
+
+        iptables_cmds = []
+
+        # Remove iptables/ip6tables commands that allow VxLAN packets
+        if ":" in self.VxlanSrcIP:
```
ZhaohuiS, line 853: @dgsudharsan Same comment as previous one.

scripts/caclmgrd
dgsudharsan: Done

scripts/caclmgrd
dgsudharsan: Done

ZhaohuiS (approved):
LGTM

dgsudharsan:
@StormLiangMS  Can you please create a new 202211 branch for sonic-host-services?

StormLiangMS:
@dgsudharsan cherry pick failed, could you submit a separate PR for 202211?

PR #54, opened 2023-04-10
[202211][caclmgrd]Allow Vxlan udp port on receiving vxlan tunnel configuration

dgsudharsan:
Manual cherry-pick of https://github.com/sonic-net/sonic-host-services/pull/50
Currently VxLAN packets are dropped due to absence of rule in iptables. When there is L3 traffic on egress VTEP with no ARP entry, due to the absence of processing by kernel, the traffic gets dropped.
Instead the VxLAN packets should be processed and ARP request needs to be resolved by the kernel.

Made changes to allow VxLAN udp port when VxLAN tunnel is configured.

Note: This change doesn't consider the VxLAN_PORT configuration in the switch table. This use case is for the VNET and for EVPN the default value of 4789 is used.

### DIFF
--- a/scripts/caclmgrd
+++ b/scripts/caclmgrd
@@ -773,14 +773,18 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
         iptables_cmds = []
 
         # Add iptables/ip6tables commands to allow VxLAN packets
-        if ":" in self.VxlanSrcIP:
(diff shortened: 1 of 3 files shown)


dgsudharsan:
Build failure due to https://github.com/sonic-net/sonic-host-services/pull/24 not present in 202211. @StormLiangMS  Can you please cherry-pick?

StormLiangMS:
@dgsudharsan I cherry-picked successfully with #50, previous conflict should be caused by some un cherry-picked PR, I think you may close this PR.

PR #57, opened 2023-04-13
[Clock] Implement clock CLI

fastiuk:
closes [#1171](https://github.com/sonic-net/SONiC/issues/1171)

### DIFF
--- a/tests/hostcfgd/hostcfgd_test.py
+++ b/tests/hostcfgd/hostcfgd_test.py
@@ -417,7 +417,8 @@ class TestHostcfgdDaemon(TestCase):
             expected = [
                 call(['sudo', 'service', 'hostname-config', 'restart']),
                 call(['sudo', 'monit', 'reload']),
-                call(['timedatectl', 'set-timezone', 'Europe/Kyiv'])
+                call(['timedatectl', 'set-timezone', 'Europe/Kyiv']),
+                call(['systemctl', 'restart', 'rsyslog']),
             ]
             mocked_subprocess.check_call.assert_has_calls(expected,
                                                           any_order=True)
--- a/tests/hostcfgd/hostcfgd_test.py
+++ b/tests/hostcfgd/hostcfgd_test.py
@@ -296,10 +296,15 @@ class TesNtpCfgd(TestCase):
 class TestHostcfgdDaemon(TestCase):
 
     def setUp(self):
+        self.get_dev_meta = mock.patch(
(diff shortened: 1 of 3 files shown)


liat-grozovik:
@fastiuk please update the PR description and the PR subject. this PR is not for CLI but for a backend.

yxieca:
The fact that this code in hostcfgd is needed, does that imply that the BIOS clock is NOT synchronized with the OS settings? Which further implies that during reboot, the clock would be wrong until hostcfgd fixes the date/time? This is not safe for upgrades.

yxieca (changes requested):
Why do we need this change?

fastiuk:
> The fact that this code in hostcfgd is needed, does that imply that the BIOS clock is NOT synchronized with the OS settings? Which further implies that during reboot, the clock would be wrong until hostcfgd fixes the date/time? This is not safe for upgrades.

No, it doesn't. The code reflected here is to update the system timezone, and it doesn't touch system time.
I would say, even more, changing the timezone is a permanent change, which will persist after a reboot, because `timedatectl` writes timezone to `/etc/timezone`.

fastiuk:
> Why do we need this change?

Which one, I don't see any code snippet

fastiuk:
Ping here

tests/hostcfgd/hostcfgd_test.py
```diff
@@ -425,31 +428,33 @@ def test_devicemeta_event(self):
...
+                    pass
 
-            expected = [
-                call(original_syslog.LOG_INFO, 'Hostname was not updated: Already set up with the same name: SameHostName')
-            ]
-            mocked_syslog.syslog.assert_has_calls(expected)
+                expected = [
+                    call(original_syslog.LOG_INFO, 'Hostname was not updated: Already set up with the same name: SameHostName')
+                ]
+                mocked_syslog.syslog.assert_has_calls(expected)
```
yxieca, line 457: Are there tabs here? If so, please change to white space. Thanks!

yxieca (approved):

yxieca:
Please confirm if there are tabs in the diff. If so, please change to white spaces.

tests/hostcfgd/hostcfgd_test.py
fastiuk: No tabs. Only spaces

fastiuk:
> Please confirm if there are tabs in the diff. If so, please change to white spaces.

no tabs in diff

PR #60, opened 2023-05-13
[NTP] Update NTP configuration via ConfigDB

fastiuk:
hld [#1219](https://github.com/sonic-net/SONiC/pull/1296)
closes [#1171](https://github.com/sonic-net/SONiC/issues/1254)

This PR brings functionality that allows us to update the NTP configs by writing to the NTP, NTP_SERVER, and NTP_KEY tables of ConfigDB.


fastiuk:
Tests will fail until https://github.com/sonic-net/sonic-swss-common/pull/781 will be merged

scripts/hostcfgd
```diff
@@ -1302,84 +1302,141 @@ class NtpCfg(object):
     1) ntp-config.service handles the configuration updates and then starts ntp.service
     2) Both of them start after all the feature services start
     3) Purpose of this daemon is to propagate runtime config changes in
-       NTP, NTP_SERVER and LOOPBACK_INTERFACE
+       NTP, NTP_SERVER, NTP_KEY, and LOOPBACK_INTERFACE
     """
+    NTP_CONF_RESTART = ['service', 'ntp-config', 'restart']
```
saiarcot895, line 1307: Why is service being used instead of systemctl?

fastiuk: No special reason. Both will work equally.

saiarcot895: Regardless, I think systemctl should be used instead of service here, since the systemctl command is already used in many other places, and this would keep things consistent.

saiarcot895 (approved):

PR #63, opened 2023-06-01
Add eth1-midplane IP Table rule only if the device type is chassis.

abdosi:
What I did:
Add eth1-midplane IP Table rule only if the device type is chassis.

Why I did:
To prevent build failure on VS/KVM testbed as the Linux command to get `eth1-midplane` ip address fails (as expected) and it logs the error message which makes test case fails because of log analyzer check.

How I verify:
UT and PR Test should cover this change.


StormLiangMS (approved):
LGTM

PR #64, opened 2023-06-05
[202211][config reload] Config Reload Enhancement

dgsudharsan:
Backporting https://github.com/sonic-net/sonic-host-services/pull/45
Code changes for config reload Enhancement https://github.com/sonic-net/SONiC/pull/1203.
Enhancing config reload to sequence the services and faster system initialization.

Immediately restart the critical services during config reload.
The non critical should be started only after all the ports are initialized.
Services can be configured to be started immediately or delayed. This can be using a field in FEATURE table.
The existing timers should be removed by this event driven approach.
This flow is applicable in case of all reboots (warm/fast/cold) as well as config reload.

### DIFF
--- a/scripts/hostcfgd
+++ b/scripts/hostcfgd
@@ -10,6 +10,7 @@ import syslog
 import signal
 import re
 import jinja2
+import threading
 from sonic_py_common import device_info
 from sonic_py_common.general import check_output_pipe
 from swsscommon.swsscommon import ConfigDBConnector, DBConnector, Table, SonicDBConfig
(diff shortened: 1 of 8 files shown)


qiluo-msft (approved):

scripts/hostcfgd
```diff
@@ -156,11 +158,15 @@ class Feature(object):
             feature_cfg (dict): Feature CONFIG_DB configuration
             deviec_config (dict): DEVICE_METADATA section of CONFIG_DB
         """
+        if 'has_timer' in feature_cfg:
```
wenyiz2021, line 161: hi @dgsudharsan , wonder why you don't want 'has_timer' in config_db.json, FEATURE section?

dgsudharsan: @wenyiz2021 Please refer to this comment https://github.com/sonic-net/sonic-host-services/pull/45#discussion_r1119395605

PR #62, opened 2023-05-31
Fix caclmgrd crash issue when applying scale cacl rules

ZhaohuiS:
Fix the issue https://github.com/sonic-net/sonic-buildimage/issues/10883.
For performance reason, libswsscommon is not thread safe by design.
caclmgrd share config DB connection cross thread, so change to use new db connector in child thread.

PR #70, opened 2023-07-08
[202205]Fix caclmgrd crash issue when applying scale cacl rules (#62)

ZhaohuiS:
Fix the issue sonic-net/sonic-buildimage#10883.
For performance reason, libswsscommon is not thread safe by design. caclmgrd share config DB connection cross thread, so change to use new db connector in child thread.


ZhaohuiS:
Close this one as https://github.com/sonic-net/sonic-buildimage/pull/15763 is merged into 202205.

PR #73, opened 2023-09-08
Update for Bookworm changes

saiarcot895:
Update Python dependency list and test code for Bookworm, and break a cyclical dependency between this repo and sonic-buildimage.

For Python package builds, newer versions of pip/setuptools don't support test_requires, and the current standard is to specify any extra dependencies (such as those required for testing) under extra_requires. Therefore, specify the testing dependencies under extra_requires. These can be installed via pip using `pip install '.[testing]'`.

In the PAM config files, replace libpam-cracklib with libpam-pwquality, since the former has been removed from Bookworm.

Finally, updating PAM config files currently causes a cyclical dependency, since the templates are in sonic-buildimage, but the test output files are in this repo. This would mean that in order to make a change in the template that would affect the output, PR checks would fail. Fix this by moving the template files into this repo. References to the template files in sonic-buildimage would need to be updated to point to the files in sonic-host-services.


saiarcot895:
Depends on #74.

saiarcot895:
@liushilongbuaa would you be able to review the pipeline-related changes here? This is removing a dependency on sonic-buildimage so that a cyclical dependency is broken.

liushilongbuaa (dismissed):
LGTM.
Take care for submodule HEAD updation.

PR #75, opened 2023-09-09
Updated the iptable rule to use parent/base name of midplane interface of chassis.

abdosi:
What I did:

Updated the iptable rule to use parent/base name of midplane interface of chassis.

Why I did:

Check here: https://github.com/sonic-net/sonic-buildimage/pull/16505


lguohan (approved):

abdosi:
@StormLiangMS  needed for 202305.

PR #81, opened 2023-10-25
TACACSPLUS_PASSKEY_ENCRYPTION support Part - II

nmoray:
- What I did
Added a support of TACACS passkey encryption feature. 
Ref. : [HLD](https://github.com/sonic-net/SONiC/pull/1471)
This PR comprises the decryption logic.

- How I did it
Implemented the feature by following [HLD](https://github.com/sonic-net/SONiC/pull/1471)

- How to verify it
`1. Configured TACACS passkey:
root@sonic:/# config tacacs passkey <pass>

2. Verified whether passkey is encrypted:
root@sonic:/# show runningconfiguration all  | grep passkey
            "passkey": "U2FsdGVkX19kFwDeP3IhgqbLJeed3pXtazJ73FtmD3I="

3. Verified /etc/pam.d/common-auth-sonic file to validate if the passkey is decrypted correctly [Referred while ssh'ing into the device]
root@sonic:~# cat /etc/pam.d/common-auth-sonic | grep secret
auth    [success=done new_authtok_reqd=done default=ignore auth_err=die]    pam_tacplus.so server=<ip>:49 secret=<pass_in_plaintext> login=login timeout=5   try_first_pass
auth    [success=done new_authtok_reqd=done default=ignore auth_err=die]    pam_tacplus.so server=<ip>:49 secret=<pass_in_plaintext> login=login timeout=5   try_first_pass

4. Verified passkey is hidden in show tacacs output
root@sonic:~# show tacacs 
TACPLUS global auth_type pap (default)
TACPLUS global timeout 5 (default)
TACPLUS global passkey configured Yes

5. Verified user able to login into device with TACACS credentials`

- Related PR(s)
[PR#3027](https://github.com/sonic-net/sonic-utilities/pull/3027)


scripts/hostcfgd
```diff
@@ -76,6 +77,38 @@ CFG_DB = "CONFIG_DB"
 STATE_DB = "STATE_DB"
 
 
+def get_salt():
+    file_path = "/etc/shadow"
+    target_username = "admin"
```
madhupalu, line 82: this variable never used, can be cleaned up

nmoray: sure

scripts/hostcfgd
```diff
@@ -500,6 +533,14 @@ class AaaCfg(object):
                 server = tacplus_global.copy()
                 server['ip'] = addr
                 server.update(self.tacplus_servers[addr])
+                if server['passkey'] is not None:
+                    config_db = ConfigDBConnector() 
+                    config_db.connect()
+                    output, errs = decrypt_passkey(server['passkey'])
```
liuh-80, line 539: What will happen when OS upgrade from old version which not support this feature? the passkey not encrypt in old version, decrypt will failed here, code here need identify this.


another solution is add code to DB migrator, but that script is very complex.

liuh-80: I noticed in the HLD: A DB migration script will be added for users to migrate existing config_db to convert tacacs passkey plaintext to encrypted.

So please also create the DB migration script PR.

nmoray: @liuh-80 As I mentioned in the HLD PR, we can achieve the backward compatibility without DB migration by simply following the logic stated above. Please share your thoughts on the same.

liuh-80: I think it's OK, please update code and HLD according to this design change.

zhangyanzhao:
Reviewers, if you are ok with this PR, please help to approve it. Thanks.

madhupalu (approved):
I am good with the changes

nmoray:
> Please resolve conflict, add testcases to coverage new code.

Yeah sure @qiluo-msft. I will do it. But unless and until https://github.com/sonic-net/sonic-buildimage/pull/17201 PR is merged, build won't be success for this PR. So if possible please help in merging the dependent PR fist.

nmoray:
@anders-nexthop @qiluo-msft @madhupalu I have closed this PR due to several merge conflicts and created a new [PR](https://github.com/sonic-net/sonic-host-services/pull/274).

PR #82, opened 2023-10-31
[DualToR][caclmgrd] Fix IPtables rules for multiple vlan interfaces for DualToR config

vdahiya12:
This PR is a required for changing the L3 IP forwarding Behavior to SoC in active-active toplogy. 
Basically a src IP is added to the SNAT rule so that only packets originating from ToR with src IP as vlan IP get natted by the rule and change the src IP to LoopBack IP
However if there are mutiple vlan IP's we only add the source IP as vlan IP,  for which the SoC IP belongs to, this PR adds that change.

#### How I did it
check the config DB if the ToR is a DualToR and has an SoC IP assigned.
put an iptable rule
iptables -t nat -A POSTROUTING --destination -j SNAT --to-source "
Signed-off-by: vaibhav-dahiya [vdahiya@microsoft.com](mailto:vdahiya@microsoft.com)

#### how to verify
```
Rules are only added once for the source IP the SoC IP belongs to

Chain PREROUTING (policy ACCEPT)
target     prot opt source               destination         

Chain INPUT (policy ACCEPT)
target     prot opt source               destination         

Chain OUTPUT (policy ACCEPT)
target     prot opt source               destination         

Chain POSTROUTING (policy ACCEPT)
target     prot opt source               destination         
SNAT       all  --  192.168.0.1          192.168.0.7          to:10.1.0.38
SNAT       all  --  192.168.1.1          192.168.1.9          to:10.1.0.38


```

### DIFF
--- a/scripts/caclmgrd
+++ b/scripts/caclmgrd
@@ -365,7 +365,7 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
 
             loopback_address = loopback_networks[0].network_address
             vlan_name = 'Vlan'
-            vlan_table = self.config_db_map[DEFAULT_NAMESPACE].get_table(self.VLAN_INTF_TABLE)
+            vlan_table = config_db_connector.get_table(self.VLAN_INTF_TABLE)
             vlan_networks = get_ipv4_networks_from_interface_table(vlan_table, vlan_name)
 
             if len(vlan_networks) == 0:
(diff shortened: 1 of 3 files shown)


lolyu (commented):
Can we add a UT for this?

zjswhhh (approved):
lgtm

StormLiangMS:
@vdahiya12 could you update ADO and test result for 202305?

StormLiangMS:
@vdahiya12  cherry pick done, pls update ADO in the PR.

PR #83, opened 2023-11-09
Disable systemd auto-restart of dependent services for spineRouters

deepak-singhal0408:
Description:
Issue details and RCA here. https://github.com/sonic-net/sonic-buildimage/issues/16991

MSFT ADO:
25422412

How I did it:
Currently featured script overrides the systemd service files of the features depending upon auto_restart enable/disable.
I am skipping dependent features(syncd, gbsyncd for now) to have "RESTART=Always"
 for them to not start immidiately, and instead get started by SWSS through swss.sh script.
The issue of syncd double stop is also applicable to pizza box platforms, however no traffic impact is seen there, where as on VOQ chassis, we do see traffic impact due to early start of syncd service.
The change can be extended to pizza box platforms as well in future once it gets vetted on SpineRouters.

How to verify it
1.  extended Unit tests to cover this scenario.
2. followed same steps as mentioned in https://github.com/sonic-net/sonic-buildimage/issues/16991 and confirmed that
a. No traffic loss end to end
b. Dont see double stop/start of syncd service

Also Ran few more tests.

Tested on both single asic and multi-asic hw platforms.
Tested config-reload
sudo reboot
service restart
syncd process crash, orchagent process crash
Confirmed from logs, that in none of the scenario, syncd gets killed twice

### DIFF
--- a/scripts/featured
+++ b/scripts/featured
@@ -319,7 +319,20 @@ class FeatureHandler(object):
         Returns:
             None.
         """
-        restart_field_str = "always" if "enabled" in feature_config.auto_restart else "no"
+        # As per the current code(due to various dependencies) SWSS service stop/start also stops/starts the dependent services(syncd, teamd, bgpd etc)
+        # There is an issue seen of syncd service getting stopped twice upon a critical process crash in syncd service due to above reason.
+        # Also early start of syncd service has traffic impact on VOQ chassis.
+        # to fix the above issue, we are disabling the auto restart of syncd service as it will be started by swss service.
+        # This change can be extended to other dependent services as well in future and also on pizza box platforms.
+
+        device_type = self._device_config.get('DEVICE_METADATA', {}).get('localhost', {}).get('type')
+        is_dependent_service = feature_config.name in ['syncd', 'gbsyncd']
+        if device_type == 'SpineRouter' and is_dependent_service:
+            syslog.syslog(syslog.LOG_INFO, "Skipped setting Restart field in systemd for {}".format(feature_config.name))
+            restart_field_str = "no"
+        else:
+            restart_field_str = "always" if "enabled" in feature_config.auto_restart else "no"
+
         feature_systemd_config = "[Service]\nRestart={}\n".format(restart_field_str)
         feature_names, feature_suffixes = self.get_multiasic_feature_instances(feature_config)
 
--- a/tests/featured/featured_test.py
+++ b/tests/featured/featured_test.py
@@ -54,7 +54,7 @@ class TestFeatureHandler(TestCase):
 
         return True if not ddiff else False
 
-    def checks_systemd_config_file(self, feature_table, feature_systemd_name_map=None):
+    def checks_systemd_config_file(self, device_type, feature_table, feature_systemd_name_map=None):
         """Checks whether the systemd configuration file of each feature was created or not
         and whether the `Restart=` field in the file is set correctly or not.
 
@@ -71,6 +71,7 @@ class TestFeatureHandler(TestCase):
                                                 'auto_restart.conf')
 
         for feature_name in feature_table:
+            is_dependent_feature = True if feature_name in ['syncd', 'gbsyncd'] else False
             auto_restart_status = feature_table[feature_name].get('auto_restart', 'disabled')
             if "enabled" in auto_restart_status:
                 auto_restart_status = "enabled"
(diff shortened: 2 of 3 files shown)


deepak-singhal0408:
@mlok-nokia @anamehra for viz and review.. thanks!

scripts/featured
```diff
@@ -319,7 +319,20 @@ class FeatureHandler(object):
...
-        restart_field_str = "always" if "enabled" in feature_config.auto_restart else "no"
+        # As per the current code(due to various dependencies) SWSS service stop/start also stops/starts the dependent services(syncd, teamd, bgpd etc)
+        # There is an issue seen of syncd service getting stopped twice upon a critical process crash in syncd service due to above reason.
+        # Also early start of syncd service has traffic impact on VOQ chassis.
+        # to fix the above issue, we are disabling the auto restart of syncd service as it will be started by swss service.
+        # This change can be extended to other dependent services as well in future and also on pizza box platforms.
+
+        device_type = self._device_config.get('DEVICE_METADATA', {}).get('localhost', {}).get('type')
+        is_dependent_service = feature_config.name in ['syncd', 'gbsyncd']
+        if device_type == 'SpineRouter' and is_dependent_service:
```
judyjoseph, line 330: We need to check if we are doing a warm boot before setting this, as we start the peer dependent services if we are doing a warm boot https://github.com/sonic-net/sonic-buildimage/blob/master/files/scripts/swss.sh#L237

abdosi: @judyjoseph as this check is for Spine route warm-boot is don;t care

abdosi (approved):
LGTM

scripts/featured
```diff
@@ -319,7 +319,20 @@ class FeatureHandler(object):
...
             None.
         """
-        restart_field_str = "always" if "enabled" in feature_config.auto_restart else "no"
+        # As per the current code(due to various dependencies) SWSS service stop/start also stops/starts the dependent services(syncd, teamd, bgpd etc)
+        # There is an issue seen of syncd service getting stopped twice upon a critical process crash in syncd service due to above reason.
+        # Also early start of syncd service has traffic impact on VOQ chassis.
+        # to fix the above issue, we are disabling the auto restart of syncd service as it will be started by swss service.
+        # This change can be extended to other dependent services as well in future and also on pizza box platforms.
+
+        device_type = self._device_config.get('DEVICE_METADATA', {}).get('localhost', {}).get('type')
```
arlakshm, line 328: nit: return None as default value for `type` ?

scripts/featured
```diff
@@ -319,7 +319,20 @@ class FeatureHandler(object):
...
+        is_dependent_service = feature_config.name in ['syncd', 'gbsyncd']
+        if device_type == 'SpineRouter' and is_dependent_service:
+            syslog.syslog(syslog.LOG_INFO, "Skipped setting Restart field in systemd for {}".format(feature_config.name))
```
arlakshm, line 331: question: This change similar to making autorestart configuration disabled for `syncd` feature for Spine devices, is that correct?.  Can we update the init_config.j2 file with condition iso adding condition here? will that work?

arlakshm (commented):
as comments

scripts/featured
deepak-singhal0408: >question: This change similar to making autorestart configuration disabled for syncd feature for Spine devices, is that correct?. Can we update the init_config.j2 file with condition iso adding condition here? will that work?

Not entirely.. We are not disabling auto restart feature completely here.. There are other side effects to that.. for eg: if we disable auto restart for syncd, any critical process crash wont exit the service and the service wont get restarted again(there is autorestart feature check in supervisor-exit-process-listener)
Code snippet:
if (process_name in critical_process_list or group_name in critical_group_list) and expected == 0:
                    is_auto_restart = get_autorestart_state(container_name)
                    if is_auto_restart != "disabled":
                        MSG_FORMAT_STR = "Process '{}' exited unexpectedly. Terminating supervisor '{}'"
                        msg = MSG_FORMAT_STR.format(payload_headers['processname'], container_name)
                        syslog.syslog(syslog.LOG_INFO, msg)
                        publish_events(events_handle, payload_headers['processname'], container_name)
                        swsscommon.events_deinit_publisher(events_handle)
                        os.kill(os.getppid(), signal.SIGTERM)
                    else:
                        process_under_alerting[process_name]["last_alerted"] = time.time()
                        process_under_alerting[process_name]["dead_minutes"] = 0

scripts/featured
deepak-singhal0408: @judyjoseph yes. once we extend this for pizza box, will add warm-boot check accordingly.

arlakshm (approved):

StormLiangMS:
@deepak-singhal0408  can't clean cherry pick to 202305.

gechiang:
@yxieca , can you help review/approve for 202205? Thanks!

yxieca:
> @yxieca , can you help review/approve for 202205? Thanks!

@gechiang in 202205 branch, this is not a submodule. A backport will have to be an PR in sonic-buildimage repo for 202205 branch.

gechiang:
Thanks @yxieca !
@deepak-singhal0408 , Please raise a PR for this change for 202205 under sonic-buildimage.

deepak-singhal0408:
Thanks @yxieca for help merging into 202205..
@gechiang here is the 202205 PR that's merged...https://github.com/sonic-net/sonic-buildimage/pull/17203

PR #85, opened 2023-11-16
[featured] Add support for dhcp_server in featured

yaqiangz:
#### Why I did it
Add support in featured for Port Based IPv4 DHCP Server, HLD: https://github.com/sonic-net/SONiC/pull/1282.
dhcp_server feature need dhcp_relay, and processes running in dhcp_relay container started by supervisord would change due to state of dhcp_server feature, hence we need to restart dhcp_relay service when dhcp_server feature is disabled/enabled.

##### Work item tracking
- Microsoft ADO **(number only)**: 25723645

#### How I did it
1. Add logic to restart dhcp_relay when state of dhcp_server feature change.
2. Add unit test

#### How to verify it
1. Unit test passed
2. Manually test in device, dhcp_relay container was restarted when state of dhcp_server feature change.

### DIFF
--- a/scripts/featured
+++ b/scripts/featured
@@ -24,15 +24,48 @@ DEFAULT_SELECT_TIMEOUT = 1000 # 1sec
 PORT_INIT_TIMEOUT_SEC = 180
 
 
-def run_cmd(cmd, log_err=True, raise_exception=False):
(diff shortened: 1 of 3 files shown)


yaqiangz:
No need

PR #86, opened 2023-11-18
Add support to make determine/process reboot-cause services restartable

anamehra:
Signed-off-by: anamehra <anamehra@cisco.com>

#### Why I did it
Fixes https://github.com/sonic-net/sonic-buildimage/issues/16990

This PR can be merged independently. The PR (https://github.com/sonic-net/sonic-buildimage/pull/17220) will need this host-services PR to be merged and released.

MSFT ADO: 25892864

1. determine-reboot-cause and process-reboot-cause service does not start If the database service fails to restart in the first attempt. Even if the Database service succeeds in next attempt, these reboot-cause services do not start.

2. The process-reboot-service does not restart if the docker or database service restarts, which leads to an empty reboot-cause history
3. deploy-mg from sonic-mgmt also triggers the docker service restart. The restart of the docker service caused the issue stated in 2 above. The docker restart also triggers determine-reboot-cause to restart which creates an additional reboot-cause file in history and modifies the last reboot-cause.

This PR along with sonic-buildimage PR (17220) fixes these issues by making both processes to start again when dependency meets after dependency failure, making both processes restart when the database service restarts, and preventing duplicate processing of the last reboot reason.

#### How I did it
1. Modified systemd unit files to make determine-reboot-cause and process-reboot-cause services restartable when the database service restarts.
2. On the restart, the determine-reboot-cause service should not recreate a new reboot-cause entry in the database. Added check for first start or restart to skip entry for restart case.

#### How to verify it
On single asic pizza box:
1.  Installed the image and check reboot-cause history
2. restart database service and verify that determine-reboot-cause and process-reboot-cause services also restart. Verify that reboot-cause shows correct data and no new entry is created for restart.

On Chassis:
1.  Installed the image and check reboot-cause history
2. restart the database service and verify that determine-reboot-cause and process-reboot-cause services also restart. Verify that reboot-cause shows correct data and no new entry is created for restart.
5. Reboot LC. On Supervicor, stop database-chassis service.
     Let database service on LC fail the first time. determine-reboot-cause and process-reboot-cause  would fail to start due to dependency failure
     start database-chassis on Supervisor. Database service on LC should now start successfully.
     Verify determine-reboot-cause and process-reboot-cause  also starts
     Verify show reboot-cause history output

PR #89, opened 2023-11-28
Revert "Add support to make determine/process reboot-cause services restartable"

anamehra:
[Reverts sonic-net/sonic-host-services#86](https://github.com/sonic-net/sonic-host-services/pull/88) fixes this issue by using Unit file settings. No need to modify script with those changes.

This change will still be required for 202205 as the change in https://github.com/sonic-net/sonic-host-services/pull/88 requires dependency changes which may not be picked in 202205


anamehra:
@gechiang , for your viz. Thanks

gechiang (approved):

PR #95, opened 2023-12-08
[caclmgrd][DualToR] Fix a case where vlan address is not network address for DualToR Active-active configuration

vdahiya12:
This PR fixes #82 where a corner case is not covered, if vlan address of DUT is not network address. 
Basically a src IP is added to the SNAT rule so that only packets originating from ToR with src IP as vlan IP get natted by the rule and change the src IP to LoopBack IP

For multiple vlan IP's only the vlan address coming from configuration is picked as the vlan address and not the network address. 

We consider a case where there are multiple vlan IP's are example in such format :

```
Vlan705                 10.131.113.17/28     up/up         N/A                   N/A
Vlan805                 10.131.111.129/26    up/up         N/A                   N/A

  "VLAN_INTERFACE": {
        "Vlan705|10.131.113.17/28": {},
        "Vlan805|10.131.111.129/26": {},
```
In such a case we expect iptables rules should be :

```
target     prot opt source               destination         
SNAT       all  --  10.131.113.17        10.131.113.19        to:10.212.64.1
SNAT       all  --  10.131.111.129       10.131.111.131       to:10.212.64.1
```

where soc_ip's are such :
```
10.131.111.131, 10.131.113.19
```

Before this change the souce IP of the vlan is not correctly picked in the iptables rule

### DIFF
--- a/scripts/caclmgrd
+++ b/scripts/caclmgrd
@@ -44,7 +44,7 @@ def _ip_prefix_in_key(key):
 
 def get_ipv4_networks_from_interface_table(table, intf_name):
     
-    addresses = []
+    addresses = {}
     if table:
         for key, _ in table.items():
             if not _ip_prefix_in_key(key):
@@ -54,8 +54,10 @@ def get_ipv4_networks_from_interface_table(table, intf_name):
             iface_name, iface_cidr = key
             if iface_name.startswith(intf_name):
                 ip_ntwrk = ipaddress.ip_network(iface_cidr, strict=False)
-                if isinstance(ip_ntwrk, ipaddress.IPv4Network):
-                    addresses.append(ip_ntwrk)
+                ip_str = iface_cidr.split("/")[0]
+                ip_addr = ipaddress.ip_address(ip_str)
+                if isinstance(ip_ntwrk, ipaddress.IPv4Network) and isinstance(ip_addr, ipaddress.IPv4Address):
+                    addresses[ip_ntwrk] = ip_addr
 
     return addresses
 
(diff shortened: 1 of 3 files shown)


scripts/caclmgrd
```diff
@@ -54,8 +54,10 @@ def get_ipv4_networks_from_interface_table(table, intf_name):
             iface_name, iface_cidr = key
             if iface_name.startswith(intf_name):
                 ip_ntwrk = ipaddress.ip_network(iface_cidr, strict=False)
-                if isinstance(ip_ntwrk, ipaddress.IPv4Network):
-                    addresses.append(ip_ntwrk)
+                ip_str = iface_cidr.split("/")[0]
```
prsunny, line 57: Could you check if this works if there are secondary IPs in Vlan? Like two or three different IPs in the same Vlan?

prsunny (commented):
Please add an example of the issue and config in the description.

zjswhhh:
> Please add an example of the issue and config in the description.

+1

scripts/caclmgrd
vdahiya12: Works as intended, added in description

vdahiya12:
> Please add an example of the issue and config in the description.

added

zjswhhh (approved):
lgtm

StormLiangMS:
@vdahiya12 could you update ADO in the description?

vdahiya12:
> @vdahiya12 could you update ADO in the description?

@StormLiangMS added

PR #80, opened 2023-10-23
Add LDAP feature support

davidpil2002:
Added LDAP support to hostcfgd

LDAP HLD: https://github.com/sonic-net/SONiC/pull/1487 

This PR depends on the following PR: https://github.com/sonic-net/sonic-buildimage/pull/16969

Details:
The fail (in the build check) reason is that this sonic-host-services LDAP unit tests have dependencies of the following PR: https://github.com/sonic-net/sonic-buildimage/pull/16969

once the PR is merged, all the tests will pass

PR #107, opened 2024-04-01
Fix issue: delayed service will not be restarted if swss crashes

Junchao-Mellanox:
**What I did**

Delayed services are counting on featured to start them when system reaches PORT INIT DONE. However, there is an extra flag check "self.is_delayed_enabled" in `port_listener`. It means that the delayed service will not be restarted if swss crashes. The PR is a fix of the issue.

**How I did**

Remove the extra flag check "self.is_delayed_enabled" in `port_listener`

**How I verified**

Manual test done, passed.

### DIFF
--- a/scripts/featured
+++ b/scripts/featured
@@ -166,7 +166,7 @@ class FeatureHandler(object):
     def port_listener(self, key, op, data):
         if not key:
             return 
-        if op == 'SET' and key == 'PortInitDone' and not self.is_delayed_enabled:
+        if op == 'SET' and key == 'PortInitDone':
             syslog.syslog(syslog.LOG_INFO, "Updating delayed features after port initialization")
             self.enable_delayed_services()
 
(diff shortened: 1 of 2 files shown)


dgsudharsan (approved):

dgsudharsan:
@qiluo-msft  Can you please help to review?

Junchao-Mellanox:
Hi @qiluo-msft , could you please help review?

qiluo-msft:
I could not understand whether it is a bug or expected that “It means that the delayed service will not be restarted if swss crashes”.
If something is a bug, could you write a unit test or sonic-mgmt testcase?

Junchao-Mellanox:
> I could not understand whether it is a bug or expected that “It means that the delayed service will not be restarted if swss crashes”. If something is a bug, could you write a unit test or sonic-mgmt testcase?

Hi @qiluo-msft , it is not a bug. It is a negative manual test. In this test, we simulate a swss crash and see that some service are down forever even if they are set to always_restart. So, the PR is to fix the restart issue.

Junchao-Mellanox:
The UT failure is caused by PR https://github.com/sonic-net/sonic-host-services/pull/80, pending on a fix or revert

qiluo-msft (approved):

Junchao-Mellanox:
Hi @yxieca, could you please help cherry-pick it?

PR #112, opened 2024-04-08
change radius authentication default

vtarnavsky:
What I did

Due to a potential security vulnerability in the Radius ciphers, we have moved to mschapv2 as it is the least vulnerable one out of the available ciphers in Radius.

How I did

changed the  RADIUS_SERVER_AUTH_TYPE_DEFAULT in hostcfgd


vtarnavsky:
@qiluo-msft could you please review?

dprital:
@qiluo-msft , Can you please review ?

scripts/hostcfgd
```diff
@@ -63,7 +63,7 @@ RADIUS_SERVER_AUTH_PORT_DEFAULT = "1812"
 RADIUS_SERVER_PASSKEY_DEFAULT = ""
 RADIUS_SERVER_RETRANSMIT_DEFAULT = "3"
 RADIUS_SERVER_TIMEOUT_DEFAULT = "5"
-RADIUS_SERVER_AUTH_TYPE_DEFAULT = "pap"
```
qiluo-msft, line 66: >pap [](http://example.com/codeflow?start=35&length=3)

Do you have another PR to change yang model? [https://github.com/sonic-net/sonic-buildimage/blob/d9e62560fdd0b67ebab99e54a0949e4684a6ee50/src/sonic-yang-models/yang-models/sonic-system-radius.yang#L70](https://github.com/sonic-net/sonic-buildimage/blob/d9e62560fdd0b67ebab99e54a0949e4684a6ee50/src/sonic-yang-models/yang-models/sonic-system-radius.yang#L70)

PR #113, opened 2024-04-10
[caclmgrd] [UT] Add unittest for caclmgrd default deny rule

lizhijianrd:
**What's the motivation of this PR?**
Add unittest for caclmgrd default deny rule:
1. Default deny rule NOT EXISTS when there is no CACL rule.
2. Default deny rule EXISTS when there is at least one CACL rules.

**How did you verify?**
Verified by run all the unittest by `python setup.py test`.

**Sign-Off By:** Zhijian Li <zhijianli@microsoft.com>

### DIFF
--- a/tests/caclmgrd/caclmgrd_default_rule_test.py
+++ b/tests/caclmgrd/caclmgrd_default_rule_test.py
@@ -1,0 +1,56 @@
+import os
+import sys
+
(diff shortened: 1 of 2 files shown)


bingwang-ms (approved):

ZhaohuiS (approved):
LGTM

PR #114, opened 2024-04-12
[caclmgrd]Fix bfd and vxlan acl rules programming in acl table update scenario

dgsudharsan:
When there is an ACL table update, caclmgrd flushes all the existing rules and recreates the rules based on configuration. This flow misses bfd and vxlan rules and thus if an ACL table is dynamically loaded, BFD and VxLAN rules would miss in the final set of rules.

Fixed the flow by checking and adding VxLAN and BFD rules

Added UT to verify the flow

### DIFF
--- a/scripts/caclmgrd
+++ b/scripts/caclmgrd
@@ -806,6 +806,12 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
             self.log_info("  " + ' '.join(cmd))
 
         self.run_commands(iptables_cmds)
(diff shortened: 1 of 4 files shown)


prsunny (approved):

dgsudharsan:
@yxieca Can you please help to cherry-pick this fix to 202311?

dgsudharsan:
@StormLiangMS Can you please cherry-pick this to 202305?

PR #120, opened 2024-04-26
Added support to render Feature Table has_global_scope field.

abdosi:
What I did:
Added support to render Feature Table `has_global_scope` based on Device
Runtime Metadata

Why I did:

In some case we want to control a given feature should run in host/global namespace based on device runtime metadata.  For example in case of chassis Linecard we want to run lldp on asic namespaces (front panel ports)  and not on host/global (eth0 might be connected internally to the Supervisor and not to external management switch)

How I verify:
Manual Veriifcation
UT updated.


abdosi:
@anamehra @mlok-nokia for viz.

scripts/featured
```diff
@@ -364,7 +365,7 @@ class FeatureHandler(object):
     def get_multiasic_feature_instances(self, feature, all_instance=False):
         # Create feature name suffix depending feature is running in host or namespace or in both
         feature_names = (
-            ([feature.name] if feature.has_global_scope or not self.is_multi_npu else []) +
+            ([feature.name] if feature.has_global_scope or all_instance or not self.is_multi_npu else []) +
```
judyjoseph, line 368: why do we need to pass  all_instance, the earlier condition should be good enough ? 
```
if feature.has_global_scope or not self.is_multi_npu
```

abdosi: @judyjoseph : for this use case `if feature.has_global_scope == False or (not self.is_multi_npu == False)`  so we want to disable this service so need to add to the service list using `all_instance`

judyjoseph (approved):
LGTM

PR #126, opened 2024-05-17
Added code to process-reboot-check that mitigates JSON decoder errors

assrinivasan:
**Why I did it**

The current process-reboot-cause script is susceptible to JSONDecodeErrors if the reboot-cause file is not a properly formatted JSON file.

**How I did it**

Added a try-except block around the actual JSON parsing bit of the code

**Microsoft ADO**: 28062468

**How I verified it**

First I verified that the issue is seen when the reboot-cause file is either an empty file or an improperly formatted JSON file:

image

image

I then replaced the existing script with the script from this PR and restarted the service:

image

As is seen, the service exited with `SUCCESS` exit code and journalctl shows the JSON Decoder exception raised when trying to parse an improper JSON file.

### DIFF
--- a/scripts/process-reboot-cause
+++ b/scripts/process-reboot-cause
@@ -52,12 +52,16 @@ def read_reboot_cause_files_and_save_state_db():
         x = TIME_SORTED_FULL_REBOOT_FILE_LIST[i]
         if os.path.isfile(x):
             with open(x, "r") as cause_file:
-                data = json.load(cause_file)
-                _hash = '{}|{}'.format(REBOOT_CAUSE_TABLE_NAME, data['gen_time'])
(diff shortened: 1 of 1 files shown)


yxieca (approved):

prgeor (approved):

gechiang:
We also need this for 202205 which T2 chassis is still actively using the branch and need this bug fix.

StormLiangMS:
hi @yxieca should cherry pick this?

yxieca:
> hi @yxieca should cherry pick this?

Yes. Let's take it to 202305 as well.

gechiang:
> > hi @yxieca should cherry pick this?
> 
> Yes. Let's take it to 202305 as well.

@yxieca 
How about 202205? can we have this as Chassis is still based on this...

PR #139, opened 2024-07-16
Fix security vulnerability in caclmgrd

ZhaohuiS:
- description:

After installation, a default iptables rule allows an attacker to bypass all others rules protecting the switch management & control-plane.

- Mitigation:

Accept only ttl-lt 2 for ICMP packets, or possibly dst-port > 1024.
Traceroute may be based on ICMP or UDP or TCP.
For ICMP: accept all icmp protocol TTL < 2 packets
For UDP: accept UDP protocol TTL < 2 and UDP dest port > 1024 packets
For TCP: accept TCP protocol TTL < 2 and TCP dest port > 1024 packets
For other mismatched packets, default action is drop

- work item:

28662516

### DIFF
--- a/scripts/caclmgrd
+++ b/scripts/caclmgrd
@@ -783,8 +783,15 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
 
         # Add iptables/ip6tables commands to allow all incoming packets with TTL of 0 or 1
         # This allows the device to respond to tools like tcptraceroute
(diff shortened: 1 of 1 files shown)


maipbui (approved):
LGTM

bingwang-ms (approved):

bingwang-ms:
LGTM. The test code in sonic-mgmt repo also need to be updated.

ZhaohuiS:
https://github.com/sonic-net/sonic-mgmt/pull/13786 test case update

PR #142, opened 2024-07-29
Add dbus service to read file stat

isabelmsft:
As part of the GNOI project (HLD: https://github.com/sonic-net/SONiC/pull/1644), we require dbus support to read file stat

This PR adds dbus support for file stat

### DIFF
--- a/host_modules/file_service.py
+++ b/host_modules/file_service.py
@@ -1,0 +1,45 @@
+"""File stat handler"""
+
+from host_modules import host_service
+import subprocess
+
+MOD_NAME = 'file'
+EXIT_FAILURE = 1
+
+import os
+
+class FileService(host_service.HostModule):
+    """
+    Dbus endpoint that executes the file command
+    """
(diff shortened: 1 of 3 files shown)


prgeor (approved):

host_modules/file_service.py
```diff
@@ -0,0 +1,45 @@
...
+EXIT_FAILURE = 1
+
+import os
+
+class FileService(host_service.HostModule):
+    """
+    Dbus endpoint that executes the file command
+    """
+    @host_service.method(host_service.bus_name(MOD_NAME), in_signature='s', out_signature='ia{ss}')
+    def get_file_stat(self, path):
```
prgeor, line 16: @isabelmsft looks like this is not used in this PR. is `get_file_stat` going to be used in future?

isabelmsft: That's right, `get_file_stat` is to be used by the File.Stat API as implemented in this PR: https://github.com/sonic-net/sonic-gnmi/pull/274/files

vvolam (approved):

PR #143, opened 2024-08-02
Support auditd configuration

maipbui:
HLD https://github.com/sonic-net/SONiC/pull/1713
Microsoft ADO **(number only)**: 27951427


scripts/hostcfgd
```diff
@@ -1170,6 +1176,49 @@ class KdumpCfg(object):
                 num_dumps = data.get("num_dumps")
             run_cmd(["sonic-kdump-config", "--num_dumps", num_dumps])
 
+
+class AuditCfg(object):
```
FengPan-Frank, line 1180: do we still need this hostcfgd change or you're going to move it into container.

PR #150, opened 2024-08-22
[BFD]Fix BFD blackout issue

dgsudharsan:
Fixes https://github.com/sonic-net/sonic-buildimage/issues/19762

When caclmgrd processes any ACL table change, this results in rules flushed and readded which was introduced in https://github.com/sonic-net/sonic-host-services/pull/114 . However in this flow the BFD and vxlan rules were added at the end which may result in traffic loss because of the DROP rule for ip2me installed. To overcome this added BFD and vxlan rules at the very start so that there is no drop seen for BFD.

Existing UT should verify the flows.

Additionally manually verified


scripts/caclmgrd
```diff
@@ -914,10 +921,15 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
             iptables_cmds.append(self.iptables_cmd_ns_prefix[namespace] +
                     ['iptables', '-I', 'INPUT', '2', '-p', 'udp', '-d', self.VxlanSrcIP, '--dport', '4789', '-j', 'ACCEPT'])
 
+        return iptables_cmds
+
+    def allow_vxlan_port(self, namespace, data):
```
prsunny, line 926: Can we add a test?

dgsudharsan: Added test for the BFD scenario

prsunny (approved):

prsunny:
@ganglyu , can you please merge this PR?

PR #157, opened 2024-08-29
sonic-host-services changes for gnoi reboot

kishanps:
Add sonic-host-services code for gnoi reboot.


kishanps:
Adding @qiluo-msft, @hdwhdw @vvolam for review

PR #171, opened 2024-10-16
DBUS API for GNOI System.SetPackage

hdwhdw:
Create image_service class for supporting GNOI implementation for SetPackage.

TESTED: unit test.

### DIFF
--- a/host_modules/image_service.py
+++ b/host_modules/image_service.py
@@ -68,7 +68,7 @@ class ImageService(host_service.HostModule):
                 for chunk in response.iter_content(chunk_size=8192):
                     tmp_file.write(chunk)
                 temp_file_path = tmp_file.name
-            os.rename(temp_file_path, save_as)
+            os.replace(temp_file_path, save_as)
             return 0, "Download successful"
         except Exception as e:
             logger.error("Failed to write downloaded image to disk: {}".format(e))
--- a/host_modules/image_service.py
+++ b/host_modules/image_service.py
@@ -37,7 +37,10 @@ class ImageService(host_service.HostModule):
              save_as: local path for the downloaded image. The directory must exist and be *all* writable.
         """
         logger.info("Download new sonic image from {} as {}".format(image_url, save_as))
-        # Check if the directory exists and has write permission.
+        # Check if the directory exists, is absolute and has write permission.
+        if not os.path.isabs(save_as):
+            logger.error("The path {} is not an absolute path".format(save_as))
+            return errno.EINVAL, "Path is not absolute"
         dir = os.path.dirname(save_as)
         if not os.path.isdir(dir):
             logger.error("Directory {} does not exist".format(dir))
--- a/host_modules/image_service.py
+++ b/host_modules/image_service.py
@@ -58,7 +58,7 @@ class ImageService(host_service.HostModule):
             where: either a local path or a remote url pointing to the image.
         """
         logger.info("Using sonic-installer to install the image at {}.".format(where))
-        cmd = ["sudo", "/usr/local/bin/sonic-installer", "install", "-y", where]
+        cmd = ["/usr/local/bin/sonic-installer", "install", "-y", where]
         result = subprocess.run(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         msg = ""
         if result.returncode:
--- a/host_modules/image_service.py
+++ b/host_modules/image_service.py
@@ -14,11 +14,11 @@ import stat
 import subprocess
 
 from host_modules import host_service
+import tempfile
 
 MOD_NAME = "image_service"
 
 DEFAULT_IMAGE_SAVE_AS = "/tmp/downloaded-sonic.bin"
-TMP_IMAGE_FILE = "/tmp/tmp-sonic_image.bin"
 
 logger = logging.getLogger(__name__)
 
@@ -60,10 +60,11 @@ class ImageService(host_service.HostModule):
                 logger.error("Failed to download image: HTTP status code {}".format(response.status_code))
                 return errno.EIO, "HTTP error: {}".format(response.status_code)
             
-            with open(TMP_IMAGE_FILE, "wb") as f:
+            with tempfile.NamedTemporaryFile(dir='/tmp',delete=False) as tmp_file:
                 for chunk in response.iter_content(chunk_size=8192):
-                    f.write(chunk)
-            os.rename(TMP_IMAGE_FILE, save_as)
+                    tmp_file.write(chunk)
+                temp_file_path = tmp_file.name
+            os.rename(temp_file_path, save_as)
             return 0, "Download successful"
         except Exception as e:
             logger.error("Failed to write downloaded image to disk: {}".format(e))
--- a/host_modules/image_service.py
+++ b/host_modules/image_service.py
@@ -1,5 +1,7 @@
 """
-Installer module for downloading and installing images.
+Services related to SONiC images, such as:
+1) Download
+2) Install
 """
 
 import logging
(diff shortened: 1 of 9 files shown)


host_modules/image_service.py
```diff
@@ -0,0 +1,67 @@
...
+            lines = result.stderr.decode().split('\n')
+            for line in lines:
+                if 'Error' in line:
+                    msg = line
+                    break
+        return result.returncode, msg
+
+
+    @host_service.method(host_service.bus_name(MOD_NAME), in_signature='s', out_signature='is')
+    def install(self, path):
```
ganglyu, line 50: Can this path be a url？

host_modules/image_service.py
```diff
@@ -0,0 +1,67 @@
...
+DEFAULT_IMAGE_SAVE_AS="/host/downloaded-sonic"
+
+logger = logging.getLogger(__name__)
+
+class ImageService(host_service.HostModule):
+    """DBus endpoint that handles downloading and installing SONiC images
+    """
+
+    @host_service.method(host_service.bus_name(MOD_NAME), in_signature='ss', out_signature='is')
+    def download(self, image_url, save_as):
```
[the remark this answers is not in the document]
hdwhdw: I think dbus services should be a thin layer above CLI and we can push the complexity of such to the caller (gNOI). So this one (download) is basically a wrapper for curl and we don't do additional logic here. What do you think?

host_modules/image_service.py
hdwhdw: Yes (didn't know that). Similar to above this is a thin layer around "sonic-installer install" so if sonic-installer supports url this also can be a url. Renamed the variable to reflect that.

host_modules/image_service.py
```diff
@@ -0,0 +1,70 @@
...
+
+import logging
+import subprocess
+import os
+
+from host_modules import host_service
+
+MOD_NAME = "image_service"
+
+DEFAULT_IMAGE_SAVE_AS = "/host/downloaded-sonic"
```
qiluo-msft, line 15: >host [](http://example.com/codeflow?start=26&length=4)

/tmp/ is more safe as a default file location.

host_modules/image_service.py
```diff
@@ -0,0 +1,70 @@
...
+logger = logging.getLogger(__name__)
+
+
+class ImageService(host_service.HostModule):
+    """DBus endpoint that handles downloading and installing SONiC images"""
+
+    @host_service.method(
+        host_service.bus_name(MOD_NAME), in_signature="ss", out_signature="is"
+    )
+    def download(self, image_url, save_as):
```
qiluo-msft, line 26: >download [](http://example.com/codeflow?start=8&length=8)

Do we need to support hash check?

host_modules/image_service.py
```diff
@@ -0,0 +1,70 @@
...
+        Args:
+             image_url: url for remote image.
+             save_as: local path for the downloaded image
+        """
+        logger.info("Download new sonic image from {} as {}".format(image_url, save_as))
+        # Create parent directory.
+        dir = os.path.dirname(save_as)
+        if not os.path.exists(dir):
+            os.makedirs(dir)
+        cmd = ["/usr/bin/curl", "-Lo", save_as, image_url]
```
qiluo-msft, line 39: >curl [](http://example.com/codeflow?start=25&length=4)

This is python code, it is better to use python library to download binary? like using requests.

qiluo-msft, line 39: >save_as [](http://example.com/codeflow?start=39&length=7)

Inspired by another gnoi API design:
target should initially write the file to a temporary location so a failure does not destroy the original file.

host_modules/image_service.py
qiluo-msft, line 26: >image_url [](http://example.com/codeflow?start=23&length=9)

always check absolute path.

qiluo-msft, line 26: >save_as [](http://example.com/codeflow?start=34&length=7)

always check absolute path.

host_modules/image_service.py
```diff
@@ -0,0 +1,70 @@
...
+
+        Args:
+             image_url: url for remote image.
+             save_as: local path for the downloaded image
+        """
+        logger.info("Download new sonic image from {} as {}".format(image_url, save_as))
+        # Create parent directory.
+        dir = os.path.dirname(save_as)
+        if not os.path.exists(dir):
+            os.makedirs(dir)
```
qiluo-msft, line 38: >makedirs [](http://example.com/codeflow?start=15&length=8)

This function will be finally called by outside, and it is an attacking point. Can we enforce that the dir must exist, and every user has the "w" permission, similar to "/tmp" permission ?

host_modules/image_service.py
```diff
@@ -0,0 +1,70 @@
...
+    )
+    def install(self, where):
+        """
+        Install a a sonic image:
+
+        Args:
+            where: either a local path or a remote url pointing to the image.
+        """
+        logger.info("Using sonic-installer to install the image at {}.".format(where))
+        cmd = ["sudo", "/usr/local/bin/sonic-installer", "install", "-y", where]
```
qiluo-msft, line 61: >sudo [](http://example.com/codeflow?start=16&length=4)

Is `sudo` needed?

hdwhdw: Interesting. I deem it necessary because it is needed for manual run, but so as many `config` command in other modules. This is not needed.

hdwhdw:
For the record: Force push to reset the email.

host_modules/image_service.py
hdwhdw: Done.

host_modules/image_service.py
hdwhdw: Done.

host_modules/image_service.py
hdwhdw: Done, and added unit tests.

host_modules/image_service.py
hdwhdw: Done.

host_modules/image_service.py
hdwhdw: Done.

host_modules/image_service.py
hdwhdw: Done.

qiluo-msft: If save_as == TMP_IMAGE_FILE, what will happen? Does it harm?

hdwhdw: Good point. I use the tempfile library instead now. The library is designed to prevent collision.

host_modules/image_service.py
```diff
@@ -0,0 +1,135 @@
...
+                        response.status_code
+                    )
+                )
+                return errno.EIO, "HTTP error: {}".format(response.status_code)
+
+            with tempfile.NamedTemporaryFile(dir="/tmp", delete=False) as tmp_file:
+                for chunk in response.iter_content(chunk_size=8192):
+                    tmp_file.write(chunk)
+                temp_file_path = tmp_file.name
+            os.rename(temp_file_path, save_as)
```
ganglyu, line 71: What will happen if this save_as already exists?

host_modules/image_service.py
```diff
@@ -0,0 +1,135 @@
...
+            hash_func = hashlib.md5()
+        else:
+            logger.error("Unsupported algorithm: {}".format(algorithm))
+            return errno.EINVAL, "Unsupported algorithm"
+
+        try:
+            with open(file_path, "rb") as f:
+                for chunk in iter(lambda: f.read(4096), b""):
+                    hash_func.update(chunk)
+            return 0, hash_func.hexdigest()
```
ganglyu, line 132: Can we use md5sum directly? Is this hash result the same as md5sum result? How long will it take to calculate checksum for a typical sonic image?

host_modules/image_service.py
hdwhdw: os.rename will do nothing. Using os.replace instead.

host_modules/image_service.py
hdwhdw: 1-2 seconds. md5sum is slightly faster but not by a noticeable margin. And since this is python code, I think we should use python library to have better control over the logic.
```
$ python3
Python 3.11.2 (main, May  2 2024, 11:59:08) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from md5 import calculate_md5
>>> calculate_md5('sonic.bin')
(2.589256763458252, '489a131693a4c98803d5a520df409bf9')
>>>
$ time md5sum sonic.bin
489a131693a4c98803d5a520df409bf9  sonic.bin

real    0m1.879s
user    0m1.621s
sys     0m0.257s

```

ganglyu (approved):

qiluo-msft (approved):

PR #173, opened 2024-10-18
Correct real time CPU Utilization calculation

FengPan-Frank:
Work item tracking
Microsoft ADO (number only):29680258

**What I did**
Correct real time CPU Utilization calculation

**Why I did it**
Now in procdockerstatsd script we iterate all processes every 2 minutes and ingest into state_db, 
https://psutil.readthedocs.io/en/latest/#psutil.cpu_percent, psutil supports both blocking and non-blocking way, and we prefer to use non-blocking way, thus now psutil.cpu_percent() used is to calculate CPU utilization since last time being triggered, current implementation will create new process object in below lines every time:
for process in psutil.process_iter(['pid', 'ppid', 'memory_percent', 'cpu_percent', 'create_time', 'cmdline']):

In order to get correct CPU utilization from last 2 minutes, we need store process object into some variable, so that every iteration will first lookup all_process_obj member and use existing process object, otherwise create new object for further use. And clean up the dead process object in case some process crashes.

**How I verified it**
Verified by UT and will check by ksoftirqd reproduced case.

### DIFF
--- a/scripts/procdockerstatsd
+++ b/scripts/procdockerstatsd
@@ -24,6 +24,7 @@ REDIS_HOSTIP = "127.0.0.1"
 
 
 class ProcDockerStats(daemon_base.DaemonBase):
+    all_process_obj = {}
 
     def __init__(self, log_identifier):
         super(ProcDockerStats, self).__init__(log_identifier)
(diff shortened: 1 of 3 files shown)


qiluo-msft:
The coverage data shows critical lines are not covered. Could you add more test coverage?

lguohan:
the description is unclear. please make clear. it is unclear to me why previous calculation is wrong, and what did you do to make it correct.

scripts/procdockerstatsd
```diff
@@ -157,6 +166,11 @@ class ProcDockerStats(daemon_base.DaemonBase):
             except (psutil.NoSuchProcess, psutil.AccessDenied, psutil.ZombieProcess):
                 pass
 
+        # erase dead process
+        for id in list(self.all_process_obj.keys()):
```
qiluo-msft, line 170: >for id in list(self.all_process_obj.keys()): [](http://example.com/codeflow?start=8&length=44)

`for id in all_process_obj:` will be just enough.

scripts/procdockerstatsd
```diff
@@ -138,10 +139,18 @@ class ProcDockerStats(daemon_base.DaemonBase):
 
     def update_processstats_command(self):
         processdata = []
-        for process in psutil.process_iter(['pid', 'ppid', 'memory_percent', 'cpu_percent', 'create_time', 'cmdline']):
+        pid_set = set()
+        for process_obj in psutil.process_iter(['pid', 'ppid', 'memory_percent', 'cpu_percent', 'create_time', 'cmdline']):
```
qiluo-msft, line 143: >process_obj [](http://example.com/codeflow?start=12&length=11)

For extreme host with too many processes, the containers will be huge and db refresh will impact performance. We may need to sort by (CPU, mem) usage, and pick first N items.

FengPan-Frank: updated

FengPan-Frank:
> > ```
> >             self.update_state_db(value, 'CMD', cmd)
> > ```
> 
> The 8 DB write could be optimized to 1.
> 
> Refers to: scripts/procdockerstatsd:196 in [aedadcf](https://github.com/sonic-net/sonic-host-services/commit/aedadcf73d316b447b6f278b64464bd2ea020220). [](commit_id = [aedadcf](https://github.com/sonic-net/sonic-host-services/commit/aedadcf73d316b447b6f278b64464bd2ea020220), deletion_comment = False)

Will fix this as separate optimization PR since this change is not cpu utilization change related.

qiluo-msft (approved):

PR #177, opened 2024-10-31
featured: use run() to run cli commands in place of check_call()

anamehra:
Signed-off-by: anamehra anamehra@cisco.com

Fixes: https://github.com/sonic-net/sonic-buildimage/issues/20662

During some reboots, it was observed that some times featured.service script command fails to start the services like pmon, snmp, lldp etc.

From logs, it was observed that 'sudo systemctl enable <service> ' command failed with errorcode 13 (SIGPIPE.

```
2024 Oct 29 01:31:26.191236 aaa14-rp INFO featured: Running cmd: '['sudo', 'systemctl', 'unmask', 'pmon.service']'
2024 Oct 29 01:31:26.211167 aaa14-rp INFO systemd[1]: Reloading.
2024 Oct 29 01:31:27.212381 aaa14-rp INFO featured: Running cmd: '['sudo', 'systemctl', 'enable', 'pmon.service']'
2024 Oct 29 01:31:27.232428 aaa14-rp INFO systemd[1]: Reloading.
2024 Oct 29 01:31:28.135667 aaa14-rp ERR featured: ['sudo', 'systemctl', 'enable', 'pmon.service'] - failed: return code - -13, output:#012None
2024 Oct 29 01:31:28.135746 aaa14-rp ERR featured: Feature 'pmon.service' failed to be enabled and started

2024 Oct 29 01:34:08.661711 aaa14-rp INFO featured: Running cmd: '['sudo', 'systemctl', 'enable', 'gnmi.service']'
2024 Oct 29 01:34:08.677242 aaa14-rp INFO systemd[1]: Reloading.
2024 Oct 29 01:34:09.316554 aaa14-rp ERR featured: ['sudo', 'systemctl', 'enable', 'gnmi.service'] - failed: return code - -13, output:#012None
2024 Oct 29 01:34:09.316791 aaa14-rp ERR featured: Feature 'gnmi.service' failed to be enabled and started
```

The issue does not recover and the pmon and other services never starts. On supervisor this also leads to swss, syncd and other related docker to stay down.

In general systemctl enable does not work for some services like pmon, snmp, lldp etc as there is no WantBy directive set for these services in unit file.

The command returns stderr :
```
"The unit files have no installation config (WantedBy=, RequiredBy=, Also=,
Alias= settings in the [Install] section, and DefaultInstance= for template
units). This means they are not meant to be enabled using systemctl.

Possible reasons for having this kind of units are:
• A unit may be statically enabled by being symlinked from another unit's
  .wants/ or .requires/ directory.
• A unit's purpose may be to act as a helper for some other unit which has
  a requirement dependency on it.
• A unit may be started when needed via activation (socket, path, timer,
  D-Bus, udev, scripted systemctl call, ...).
• In case of template units, the unit is meant to be enabled with some
  instance name specified.
”
```
featured python script uses subprocess.check_call() function to invoke the command which looks like is not very reliable at handling the stderr and may cause SIGPIPE with big buffer data.

Modifying the function to use subprocess.run() resolves this issue.

run() is more reliable at handing the return data.

Validated the change with multiple reboots.


anamehra:
Hi @abdosi , please review.

anamehra:
Hi @judyjoseph , please help with this PR review. Thanks

scripts/featured
```diff
@@ -26,7 +26,11 @@ PORT_INIT_TIMEOUT_SEC = 180
 
 def run_cmd(cmd, log_err=True, raise_exception=False):
     try:
-        subprocess.check_call(cmd)
+        result = subprocess.run(cmd,
```
judyjoseph, line 29: Using subprocess.run() i.s.o subprocess.check_call() lgtm, btw are we ok to dump the result.stdout, result.stderr in the syslog as INFO - will that be huge data 

Also it will be good to see why/how we got into this condition of "systemctl pmon enable" failing. Are we now trying to enable this service early via featured ?

anamehra: The stderr/stdout should not be huge but useful info returned by systemctl command to help any failure. In worst case we see the dump I listed in description when enable fail due to the service which does not have a WantBy directive.

systemctl pmon enable fails (and so do snmp, gnmi, lldp, etc) as these service can't not be enabled by systemctl as there is no WantBy directive. So systemctl just return the message suggesting same and does nothing. Its a no op.
These services are now started by featured using 'systemctl start <>' command each boot.

anamehra: pmon used to have WantedBy directive which was removed when pmon.timer was added by https://github.com/sonic-net/sonic-buildimage/commit/4ec3af86af5c9ed2da21df33111fa84181069a3e
After this timer used to trigger pmon.
Later timer was removed to introduce featured
so pmon is not enabled any more for systemd to start it.
featured starts it with 'systemctl start'
'systemctl enable' is a no op here now

judyjoseph (approved):
Lgtm

qiluo-msft (approved):

anamehra:
Hi @yejianquan , please help to get this one in 202405. Thanks

yejianquan:
> Hi @yejianquan , please help to get this one in 202405. Thanks

@bingwang-ms , could you help to cherry-pick? Thanks!

PR #179, opened 2024-11-05
DBUS API for Containerz.StopContainer

hdwhdw:
DBUS API for Containerz.StopContainer


host_modules/docker_service.py
```diff
@@ -0,0 +1,86 @@
...
+
+from host_modules import host_service
+import docker
+import signal
+import errno
+
+MOD_NAME = "docker_service"
+
+
+class DockerService(host_service.HostModule):
```
qiluo-msft, line 11: >DockerService [](http://example.com/codeflow?start=6&length=13)

Could you reuse the code inside the SystemdService class?

hdwhdw: We could do that. But eventually we still need a docker service for upgrade because even if we can stop the old container using systemd, launching new container with new image:tag still requires a `docker run` command (correct me if I am wrong) which still requires this `docker service`. 

If that's so we might as well let this relatively code in and build on top of it.

qiluo-msft: Agree with docker service vs systemd service. I feel there is value in ALLOWED_SERVICES. Could you reuse that array or implemnet a similar ALLOWED_CONTAINERS.

hdwhdw: Done.

qiluo-msft (approved):

PR #191, opened 2024-11-29

rkavitha-hcl:
Adding sonic-host-services changes for warm reboot .
Adding HALT method support for sonic-host-services

### DIFF
--- a/host_modules/reboot.py
+++ b/host_modules/reboot.py
@@ -1,0 +1,152 @@
+"""reboot module which performs reboot"""
+
+import json
+import logging
+import threading
+import time
+from host_modules import host_service
+from utils.run_cmd import _run_command
+
+MOD_NAME = 'reboot'
+# Reboot method in reboot request
+# Both enum and string representations are supported
+REBOOTMETHOD_COLD_BOOT_VALUES = {1, "COLD"}
+REBOOTMETHOD_WARM_BOOT_VALUES = {4, "WARM"}
+REBOOTMETHOD_NSF_VALUES = {5, "NSF"}
+
+# Timeout for SONiC Host Service to be killed during reboot
+REBOOT_TIMEOUT = 260
+
+EXECUTE_COLD_REBOOT_COMMAND = "sudo reboot"
+EXECUTE_NSF_REBOOT_COMMAND = "/etc/init.d/gpins-nsf-boot nsf-reboot"
+
+logger = logging.getLogger(__name__)
+
+
+class Reboot(host_service.HostModule):
+    """DBus endpoint that executes the reboot and returns the reboot status
+    """
+
+    def __init__(self, mod_name):
+        """Use threading.lock mechanism to read/write into response_data
+           since response_data can be read/write by multiple threads"""
+        self.lock = threading.Lock()
+        # reboot_status_flag is used to keep track of reboot status on host
+        self.reboot_status_flag = {}
+        # Populating with default value i.e., no active reboot
+        self.populate_reboot_status_flag()
+        super(Reboot, self).__init__(mod_name)
+
+    def populate_reboot_status_flag(self, active = False, when = 0, reason = ""):
+        """Populates the reboot_status_flag with given input params"""
+        self.lock.acquire()
+        self.reboot_status_flag["active"] = active
+        self.reboot_status_flag["when"] = when
+        self.reboot_status_flag["reason"] = reason
+        self.lock.release()
+        return
+
+    def validate_reboot_request(self, reboot_request):
+        # Check whether reboot method is present.
+        if "method" not in reboot_request:
+            return 1, "Reboot request must contain a reboot method"
+
+        # Check whether reboot method is valid.
+        rebootmethod = reboot_request["method"]
+        valid_method = False
+        for values in [REBOOTMETHOD_COLD_BOOT_VALUES, REBOOTMETHOD_NSF_VALUES]:
+            if rebootmethod in values:
+                valid_method = True
+        if not valid_method:
+            return 1, "Invalid reboot method: " + str(rebootmethod)
+
+        # Check whether delay is non-zero. delay key will not exist in reboot_request if it is zero
+        if "delay" in reboot_request and reboot_request["delay"] != 0:
+            return 1, "Delayed reboot is not supported"
+        return 0, ""
+
+    def execute_reboot(self, rebootmethod):
+        """Execute reboot and reset reboot_status_flag when reboot fails"""
+
+        if rebootmethod in REBOOTMETHOD_COLD_BOOT_VALUES:
+            command = EXECUTE_COLD_REBOOT_COMMAND
+            logger.warning("%s: Issuing cold reboot", MOD_NAME)
+        elif rebootmethod in REBOOTMETHOD_NSF_VALUES:
+            command = EXECUTE_NSF_REBOOT_COMMAND
+            logger.warning("%s: Issuing NSF reboot", MOD_NAME)
+        else:
+            logger.error("%s: Invalid reboot method: %d", MOD_NAME, rebootmethod)
+            return
+
+        rc, stdout, stderr = _run_command(command)
+        if rc:
+            self.populate_reboot_status_flag()
+            logger.error("%s: Reboot failed execution with stdout: %s, "
+                         "stderr: %s", MOD_NAME, stdout, stderr)
+            return
+
+        """Wait for 260 seconds for the reboot to complete. Here, we expect that SONiC Host Service
+           will be killed during this waiting period if the reboot is successful. If this module
+           is still alive after the below waiting period, we can conclude that the reboot has failed.
+           Each container can take up to 20 seconds to get killed. In total, there are 10 containers,
+           and adding a buffer of 1 minute brings up the delay value to be 260 seconds."""
+        time.sleep(REBOOT_TIMEOUT)
+        # Conclude that the reboot has failed if we reach this point
+        self.populate_reboot_status_flag()
+        return
+
+    @host_service.method(host_service.bus_name(MOD_NAME), in_signature='as', out_signature='is')
+    def issue_reboot(self, options):
+        """Issues reboot after performing the following steps sequentially:
+          1. Checks that reboot_status_flag is not set
+          2. Validates the reboot request
+          3. Sets the reboot_status_flag
+          4. Issues the reboot in a separate thread
+        """
+        logger.warning("%s: issue_reboot rpc called", MOD_NAME)
+        self.lock.acquire()
+        is_reboot_ongoing = self.reboot_status_flag["active"]
+        self.lock.release()
+        # Return without issuing the reboot if the previous reboot is ongoing
+        if is_reboot_ongoing:
+            return 1, "Previous reboot is ongoing"
+
+        """Convert input json formatted reboot request into python dict.
+           reboot_request is a python dict with the following keys:
+               method - specifies the method of reboot
+               delay - delay to issue reboot, key exists only if it is non-zero
+               message - reason for reboot
+               force - either true/false, key exists only if it is true
+        """
+        try:
+            reboot_request = json.loads(options[0])
+        except ValueError:
+            return 1, "Failed to parse json formatted reboot request into python dict"
+
+        # Validate reboot request
+        err, errstr = self.validate_reboot_request(reboot_request)
+        if err:
+            return err, errstr
+
+        # Sets reboot_status_flag to be in active state
+        self.populate_reboot_status_flag(True, int(time.time()), reboot_request["message"])
+
+        # Issue reboot in a new thread and reset the reboot_status_flag if the reboot fails
+        try:
+            t = threading.Thread(target=self.execute_reboot, args=(reboot_request["method"],))
+            t.start()
+        except RuntimeError as error:
+            return 1, "Failed to start thread to execute reboot with error: " + str(error)
+        return 0, "Successfully issued reboot"
+
+    @host_service.method(host_service.bus_name(MOD_NAME), in_signature='', out_signature='is')
+    def get_reboot_status(self):
+        """Returns current reboot status on host in json format"""
+        self.lock.acquire()
+        response_data = json.dumps(self.reboot_status_flag)
+        self.lock.release()
+        return 0, response_data
+
+def register():
+    """Return the class name"""
+    return Reboot, MOD_NAME
--- a/host_modules/reboot.py
+++ b/host_modules/reboot.py
@@ -4,21 +4,25 @@ import json
 import logging
 import threading
 import time
+import docker
 from host_modules import host_service
 from utils.run_cmd import _run_command
 
 MOD_NAME = 'reboot'
 # Reboot method in reboot request
 # Both enum and string representations are supported
-REBOOTMETHOD_COLD_BOOT_VALUES = {1, "COLD"}
-REBOOTMETHOD_WARM_BOOT_VALUES = {4, "WARM"}
-REBOOTMETHOD_NSF_VALUES = {5, "NSF"}
+# Define an Enum for Reboot Methods which are defined as in https://github.com/openconfig/gnoi/blob/main/system/system.pb.go#L27
+REBOOT_METHOD_COLD_BOOT_VALUES = {1, "COLD"}
+REBOOT_METHOD_HALT_BOOT_VALUES = {3, "HALT"}
+REBOOT_METHOD_WARM_BOOT_VALUES = {4, "WARM"}
 
 # Timeout for SONiC Host Service to be killed during reboot
 REBOOT_TIMEOUT = 260
+HALT_TIMEOUT = 60
 
 EXECUTE_COLD_REBOOT_COMMAND = "sudo reboot"
-EXECUTE_NSF_REBOOT_COMMAND = "/etc/init.d/gpins-nsf-boot nsf-reboot"
+EXECUTE_HALT_REBOOT_COMMAND = "sudo reboot -p"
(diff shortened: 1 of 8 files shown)


host_modules/gnoi_reboot.py
```diff
@@ -12,13 +12,12 @@
...
 REBOOTMETHOD_COLD_BOOT_VALUES = {1, "COLD"}
 REBOOTMETHOD_WARM_BOOT_VALUES = {4, "WARM"}
-REBOOTMETHOD_NSF_VALUES = {5, "NSF"}
 
 # Timeout for SONiC Host Service to be killed during reboot
 REBOOT_TIMEOUT = 260
 
 EXECUTE_COLD_REBOOT_COMMAND = "sudo reboot"
-EXECUTE_NSF_REBOOT_COMMAND = "/etc/init.d/gpins-nsf-boot nsf-reboot"
+EXECUTE_WARM_REBOOT_COMMAND = "/usr/local/bin/warm-reboot -v"
```
kishanps, line 20: @hdwhdw @qiluo-msft Can you pls confirm this is the right command to invoke warm reboot in SONiC ?

tests/gnoi_reboot_test.py
```diff
@@ -23,13 +23,14 @@
 
 REBOOTMETHOD_UNKNOWN_ENUM = 0
 REBOOTMETHOD_COLD_BOOT_ENUM = 1
-REBOOTMETHOD_NSF_ENUM = 5
+REBOOTMETHOD_WARM_BOOT_ENUM = 4
+#REBOOTMETHOD_NSF_ENUM = 6
```
kishanps, line 27: Do these enums have a dependency on some other values ? Why is 5 changed to 4 ? And can we remove the commented line ?

kishanps:
@github76543 Joh, can you PTAL and signoff.

tests/gnoi_reboot_test.py
rkavitha-hcl: @kishanps, In gnoi (https://github.com/openconfig/gnoi/blob/main/system/system.proto), Warm reboot is set to 4.  So, I have set to 4.

host_modules/gnoi_reboot.py
hdwhdw: Yess. https://github.com/sonic-net/sonic-utilities/blob/34d2c8ce007b077b75a29ade8f0876374f66f21f/scripts/warm-reboot#L1

hdwhdw (changes requested):
Can you please look at https://github.com/sonic-net/sonic-host-services/commit/ff73070d7def8689bc92dd0e28d3dab5d0254450 and see if this is something you can reuse?

kishanps:
> Can you please look at [ff73070](https://github.com/sonic-net/sonic-host-services/commit/ff73070d7def8689bc92dd0e28d3dab5d0254450) and see if this is something you can reuse?

Thanks @hdwhdw for the reference. The reboot dbus service also needs a request/response framework which is what this PR does and IIUC @vvolam went with the other one as a stop gap solution.

Adding @github76543 (John) for additional inputs.

hdwhdw:
@kishanps thanks for clarifying. If so consider renaming the service to something more general than gnoi_reboot. Maybe 'async_system'? Having one module for each gnoi service can clutter the dbus codebase.

Also does it make sense to add your api to systemd service and call it async reboot, alongside @vvolam API?

kishanps:
> @kishanps thanks for clarifying. If so consider renaming the service to something more general than gnoi_reboot. Maybe 'async_system'? Having one module for each gnoi service can clutter the dbus codebase.
> 
> Also does it make sense to add your api to systemd service and call it async reboot, alongside @vvolam API?

@hdwhdw @github76543  @rkavitha-hcl @jaanah-hcl

I discussed with John, Dawei Huang & @vvolam and we all agree that reboot will be a separate dbus service and hence rename gnoi_reboot to just reboot. And remove the commit id ff73070 alongwith this PR to avoid the duplication.

host_modules/reboot.py
```diff
@@ -0,0 +1,151 @@
...
+# Reboot method in reboot request
+# Both enum and string representations are supported
+REBOOTMETHOD_COLD_BOOT_VALUES = {1, "COLD"}
+REBOOTMETHOD_WARM_BOOT_VALUES = {4, "WARM"}
+
+# Timeout for SONiC Host Service to be killed during reboot
+REBOOT_TIMEOUT = 260
+
+EXECUTE_COLD_REBOOT_COMMAND = "sudo reboot"
+EXECUTE_WARM_REBOOT_COMMAND = "/usr/local/bin/warm-reboot -v"
```
kishanps, line 20: @vvolam @hdwhdw @qiluo-msft 
Can you pls confirm if the script used here is the correct one for warm reboot ?

host_modules/reboot.py
```diff
@@ -0,0 +1,149 @@
...
+        return
+
+    def validate_reboot_request(self, reboot_request):
+        # Check whether reboot method is present.
+        if "method" not in reboot_request:
+            return 1, "Reboot request must contain a reboot method"
+
+        # Check whether reboot method is valid.
+        reboot_method = reboot_request["method"]
+        valid_method = False
```
vvolam, line 55: This variable does not seem to be used.

host_modules/reboot.py
vvolam: We can directly use warm-reboot command directly instead.

host_modules/gnoi_reboot.py
vvolam: @rkavitha-hcl @kishanps please use the command "sudo warm-reboot" directly instead of the path to warm-reboot as discussed in our last meeting. command is mapped to path variable in the sonic and mentioning the path is not required.

host_modules/reboot.py
rkavitha-hcl: > We can directly use warm-reboot command directly instead.

warm reboot command changed to warm-reboot

host_modules/reboot.py
rkavitha-hcl: > This variable does not seem to be used.
unused variable removed

host_modules/gnoi_reboot.py
rkavitha-hcl: > @rkavitha-hcl @kishanps please use the command "sudo warm-reboot" directly instead of the path to warm-reboot as discussed in our last meeting. command is mapped to path variable in the sonic and mentioning the path is not required.

Command is replaced with "sudo warm-reboot"

kishanps:
@qiluo-msft For review and approval

bibhuprasad-hcl:
@vvolam , could you please review

bibhuprasad-hcl:
@hdwhdw , could you please review.

scripts/sonic-host-server
```diff
@@ -21,10 +21,9 @@ def register_dbus():
             'config': config_engine.Config('config'),
             'gcu': gcu.GCU('gcu'),
             'host_service': host_service.HostService('host_service'),
+	    'reboot': reboot.Reboot('reboot'),
             'showtech': showtech.Showtech('showtech'),
             'systemd': systemd_service.SystemdService('systemd'),
-            'image_service': image_service.ImageService('image_service'),
-            'docker_service': docker_service.DockerService('docker_service'),
```
hdwhdw, line 27: Is this an accident during rebase?

hdwhdw (changes requested):

host_modules/reboot.py
```diff
@@ -0,0 +1,154 @@
...
+            return
+
+        rc, stdout, stderr = _run_command(command)
+        if rc:
+            self.populate_reboot_status_flag()
+            logger.error("%s: Reboot failed execution with stdout: %s, "
+                         "stderr: %s", MOD_NAME, stdout, stderr)
+            return
+
+        """Wait for 260 seconds for the reboot to complete. Here, we expect that SONiC Host Service
```
hdwhdw, line 90: Nits: the number 260 is not hardcoded but injected using a constant. Consider removing it from the documentation. It can change in the future.

host_modules/reboot.py
```diff
@@ -0,0 +1,154 @@
...
+           is still alive after the below waiting period, we can conclude that the reboot has failed.
+           Each container can take up to 20 seconds to get killed. In total, there are 10 containers,
+           and adding a buffer of 1 minute brings up the delay value to be 260 seconds."""
+        time.sleep(REBOOT_TIMEOUT)
+        # Conclude that the reboot has failed if we reach this point
+        self.populate_reboot_status_flag()
+        return
+
+    @host_service.method(host_service.bus_name(MOD_NAME), in_signature='as', out_signature='is')
+    def issue_reboot(self, options):
```
hdwhdw, line 101: Can you add some more documentation in what is the difference between issue_reboot and execute_reboot?

scripts/sonic-host-server
rkavitha-hcl: > Is this an accident during rebase?

These two lines were removed during code merge with Halt reboot PR. It's added back.

host_modules/reboot.py
rkavitha-hcl: > Nits: the number 260 is not hardcoded but injected using a constant. Consider removing it from the documentation. It can change in the future.

constant 260 is removed.

host_modules/reboot.py
rkavitha-hcl: > Can you add some more documentation in what is the difference between issue_reboot and execute_reboot?

documentation is added .

host_modules/reboot.py
```diff
@@ -0,0 +1,157 @@
...
+            command = EXECUTE_COLD_REBOOT_COMMAND
+            logger.warning("%s: Issuing cold reboot", MOD_NAME)
+        elif reboot_method in REBOOT_METHOD_HALT_BOOT_VALUES:
+            command = EXECUTE_HALT_REBOOT_COMMAND
+            logger.warning("%s: Issuing halt reboot", MOD_NAME)
+        elif reboot_method in REBOOT_METHOD_WARM_BOOT_VALUES:
+            command = EXECUTE_WARM_REBOOT_COMMAND
+            logger.warning("%s: Issuing WARM reboot", MOD_NAME)
+        else:
+            logger.error("%s: Invalid reboot method: %d", MOD_NAME, reboot_method)
```
vvolam, line 81: Same here. Instead of Invalid, can we use Unsupported to be clear?

host_modules/reboot.py
```diff
@@ -0,0 +1,157 @@
...
+            return
+
+        rc, stdout, stderr = _run_command(command)
+        if rc:
+            self.populate_reboot_status_flag()
+            logger.error("%s: Reboot failed execution with stdout: %s, "
+                         "stderr: %s", MOD_NAME, stdout, stderr)
+            return
+
+        """Wait for the reboot to complete. Here, we expect that SONiC Host Service
```
vvolam, line 91: I just realized this check!!

This check is wrong for HALT method, in case of HALT, gnmi container will be alive and only pmon and syncd containers will be killed for now. Can we modify the logic accordingly to below logic?

Just wait for 30 or 60 secs timeout and if pmon container is not killed, HALT method is failed.

hdwhdw (approved):
Please wait until the approval of @vvolam as well.

host_modules/reboot.py
kishanps: @vvolam I suppose the reboot -p will keep gnmi and bring down all other containers. In that case, should we just ensure that gnmi is the only container up in case of HALT ? Maybe keep the timeout the same to take care of other use cases where more containers need to be brought down.

vvolam: @kishanps Yes, we can do that because we always ensure gnmi container is up in case of HALT.

host_modules/reboot.py
bibhuprasad-hcl: > Same here. Instead of Invalid, can we use Unsupported to be clear?

string is changed to "Unsupported reboot method"

rkavitha-hcl: > Same here. Instead of Invalid, can we use Unsupported to be clear?

String is changed to  "Unsupported reboot method".

host_modules/reboot.py
```diff
@@ -0,0 +1,157 @@
...
+    def validate_reboot_request(self, reboot_request):
+        # Check whether reboot method is present.
+        if "method" not in reboot_request:
+            return 1, "Reboot request must contain a reboot method"
+
+        # Check whether reboot method is valid.
+        reboot_method = reboot_request["method"]
+        valid_reboot_method = REBOOT_METHOD_COLD_BOOT_VALUES | REBOOT_METHOD_HALT_BOOT_VALUES | REBOOT_METHOD_WARM_BOOT_VALUES
+        if reboot_method not in valid_reboot_method:
+            return 1, "Invalid reboot method: " + str(reboot_method)
```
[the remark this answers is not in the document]
rkavitha-hcl: > Instead of Invalid, can we make the string as "Unsupported reboot method"?

String is changed to  "Unsupported reboot method".

host_modules/reboot.py
vvolam: @rkavitha-hcl could you fix this check for HALT case, as the gnmi container will still be running?

kishanps: Kavitha is out sick, @jaanah-hcl Can you pls take care of this.

vvolam:
@kishanps @rkavitha-hcl Could you fix build failures as well?

rkavitha-hcl:
> @kishanps @rkavitha-hcl Could you fix build failures as well?

Build is fixed and branch is rebased

host_modules/reboot.py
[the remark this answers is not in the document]
bibhuprasad-hcl: > Instead of Invalid, can we make the string as "Unsupported reboot method"?

String is replaced with string "Unsupported reboot method"

host_modules/reboot.py
bibhuprasad-hcl: > Same here. Instead of Invalid, can we use Unsupported to be clear?

String is replaced with string "Unsupported reboot method"

host_modules/reboot.py
kishanps: @vvolam One more question on reboot -p, does it keep framework container also up along with gnmi container ? Don't you need framework container for the reboot status after the HALT execution ?

vvolam: @kishanps As I mentioned, it only kills syncd and pmon containers as of now and all remaining containers stay up.

kishanps: There are 2 parts here 
- What are the containers you want to kill ? For smartswitch, pmon & syncd may suffice but if its a regular switch you probably need to kill other running containers also. In which case, you may want to kill all containers except gnmi & framework (tied to the next part)
- Do you need framework container after HALT to query reboot status ? I don't know the use-case of HALT, so you may be the better person to call on that. If you intend to make reboot status call, then you need framework container also to give you back the status.

vvolam: @kishanps HALT script changes in reboot command are same across all switches. We are only killing syncd and pmon containers and all other containers remain alive.

host_modules/reboot.py
```diff
@@ -0,0 +1,188 @@
...
+           Each container can take up to 20 seconds to get killed. In total, there are 10 containers,
+           and adding a buffer of 1 minute brings up the delay value.
+           For Halt reboot_method, wait for 60 secs timeout. we expect pmon, syncd containers are killed, 
+           if Halt reboot is Successful."""
+        if reboot_method in REBOOT_METHOD_HALT_BOOT_VALUES:
+           try:
+               time.sleep(HALT_TIMEOUT)
+               pmon_stopped = not is_container_running("pmon")
+               syncd_stopped = not is_container_running("syncd")
+               if not (pmon_stopped and syncd_stopped):
```
vvolam, line 119: Fail HALT method only if pmon container is not stopped. pmon is the only critical container which should not be running when services are halted. syncd won't be killed in certain hardware skus, so please don't check for that container.

host_modules/reboot.py
```diff
@@ -0,0 +1,188 @@
...
+                   logger.error("%s: HALT reboot failed: pmon or syncd is still running", MOD_NAME)
+                   self.populate_reboot_status_flag()
+                   return
+           except Exception as e:
+               logger.error("%s: Error received when checking containter status for pmon or syncd [%s]:", MOD_NAME, str(e))
+
+        else:
+            time.sleep(REBOOT_TIMEOUT)
+        # Conclude that the reboot has failed if we reach this point
```
vvolam, line 128: This line is not clear to me. So we have to conclude this to be failed only for COLD and WARM reboot methods right?

vvolam (commented):
Other than those minor changes, remaining LGTM. Thank you.

host_modules/reboot.py
```diff
@@ -0,0 +1,187 @@
...
+           is still alive after the below waiting period, we can conclude that the reboot has failed.
+           Each container can take up to 20 seconds to get killed. In total, there are 10 containers,
+           and adding a buffer of 1 minute brings up the delay value.
+           For Halt reboot_method, wait for 60 secs timeout. we expect pmon, syncd containers are killed, 
+           if Halt reboot is Successful."""
+        if reboot_method in REBOOT_METHOD_HALT_BOOT_VALUES:
+            time.sleep(HALT_TIMEOUT)
+            is_pmon_running = self.is_container_running("pmon")
+            if is_pmon_running:
+                #Halt reboot has failed, as pnom is still running.
```
vvolam, line 118: nit: typo "pnom"

vvolam (approved):

host_modules/reboot.py
bibhuprasad-hcl: > nit: typo "pnom"

typo corrected.

PR #193, opened 2024-12-06
Update pipeline to Bookworm

saiarcot895:


FengPan-Frank (approved):
Thanks for the change.

PR #195, opened 2024-12-16
Adding support for filtering by destination IP and/or incoming interface to caclmgrd.

nyanto:
This feature - filtering by destination IP and/or incoming interface - can be useful to limit certain traffic from reaching the control plane via unwanted interfaces, such as VLAN interfaces or P2Ps.

### DIFF
--- a/scripts/caclmgrd
+++ b/scripts/caclmgrd
@@ -768,6 +768,14 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
                                 if rule_props["PACKET_ACTION"] == "ACCEPT":
                                     ipv4_src_ip_set.add(rule_props["SRC_IP"])
 
+                            if "DST_IPV6" in rule_props and rule_props["DST_IPV6"]:
+                                rule_cmd += ["-d", str(rule_props["DST_IPV6"])]
+                            elif "DST_IP" in rule_props and rule_props["DST_IP"]:
+                                rule_cmd += ["-d", str(rule_props["DST_IP"])]
+
+                            if "INC_INTF" in rule_props and rule_props["INC_INTF"]:
+                                rule_cmd += ["-i", str(rule_props["INC_INTF"])]
+
                             # Destination port 0 is reserved/unused port, so, using it to apply the rule to all ports.
                             if dst_port != "0":
                                 rule_cmd += ["--dport", str(dst_port)]
--- a/scripts/caclmgrd
+++ b/scripts/caclmgrd
@@ -773,8 +773,8 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
                             elif "DST_IP" in rule_props and rule_props["DST_IP"]:
                                 rule_cmd += ["-d", str(rule_props["DST_IP"])]
 
-                            if "INC_INTF" in rule_props and rule_props["INC_INTF"]:
-                                rule_cmd += ["-i", str(rule_props["INC_INTF"])]
+                            if "IN_PORTS" in rule_props and rule_props["IN_PORTS"]:
+                                rule_cmd += ["-i", str(rule_props["IN_PORTS"])]
 
                             # Destination port 0 is reserved/unused port, so, using it to apply the rule to all ports.
                             if dst_port != "0":
--- a/tests/caclmgrd/test_external_client_acl_vectors.py
+++ b/tests/caclmgrd/test_external_client_acl_vectors.py
@@ -139,7 +139,7 @@ EXTERNAL_CLIENT_ACL_TEST_VECTOR = [
                         "PACKET_ACTION": "ACCEPT",
                         "PRIORITY": "9998",
                         "DST_IP": "0.0.0.0/0",
-                        "INC_INTF": "mgmt"
+                        "IN_PORTS": "mgmt"
                     },
                 },
                 "DEVICE_METADATA": {
--- a/tests/caclmgrd/test_external_client_acl_vectors.py
+++ b/tests/caclmgrd/test_external_client_acl_vectors.py
@@ -78,6 +78,83 @@ EXTERNAL_CLIENT_ACL_TEST_VECTOR = [
         }
     ],
     [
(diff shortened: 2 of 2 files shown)


nyanto:
Hi @bingwang-ms , I noticed you reviewed a recent PR related to caclmgrd. Would you be able to assist with reviewing this one, or point me to the appropriate contact? Thanks.

bingwang-ms:
@ZhaohuiS Could you help review?

scripts/caclmgrd
```diff
@@ -768,6 +768,14 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
                                 if rule_props["PACKET_ACTION"] == "ACCEPT":
                                     ipv4_src_ip_set.add(rule_props["SRC_IP"])
 
+                            if "DST_IPV6" in rule_props and rule_props["DST_IPV6"]:
+                                rule_cmd += ["-d", str(rule_props["DST_IPV6"])]
+                            elif "DST_IP" in rule_props and rule_props["DST_IP"]:
+                                rule_cmd += ["-d", str(rule_props["DST_IP"])]
+
+                            if "INC_INTF" in rule_props and rule_props["INC_INTF"]:
```
ZhaohuiS, line 776: @nyanto does `acl_loader` command support `INC_INTF`? Could you please provide the test evidence that `acl_loader` can load the acl.json which has INC_INTF inside successfully?

nyanto: @ZhaohuiS some of the supplied unit tests contain "INC_INTF" and are passing. 
acl_loader is able to show INC_INTF successfully, see below. 

`root@device:~# acl-loader show rule
Table                Rule          Priority    Action    Match
ACLSSH               RULE_20       20          ACCEPT    INC_INTF: mgmt
                                                         L4_DST_PORT: 22
                                                         SRC_IP: 0.0.0.0/0`

If your question is related to loading an openconfig format ACL via acl-loader, then I see your point. 
It looks like acl_loader calls this "IN_PORTS":

            rule_props["IN_PORTS"] = rule.input_interface.interface_ref.config.interface

However, I do not see "IN_PORTS" used in caclmgrd or elsewhere. I could change "INC_INTF" to "IN_PORTS" if this looks like it would make it consistent with the existing schema.

ZhaohuiS: @nyanto I mean how do you load the config into config_db.json? are you using `acl-loader` to load the INC_INTF these acl configurations into config_db.json?
`sudo acl-loader update full /tmp/acl.json`

I didn't find `INC_INTF` in https://github.com/sonic-net/sonic-utilities/blob/master/acl_loader/main.py, so I am worried if acl-loader support your new config.

nyanto: @ZhaohuiS so the way we load the config is ```config load -y acl.json```
In looking at acl_loader, I do not see that it ever looks at dict keys, except while converting from openconfig format. If I try your command, it errors out because the json is not openconfig format:
```  File "/usr/local/lib/python3.9/dist-packages/acl_loader/main.py", line 334, in parse_acl_json
    yang_acl = pybindJSON.load(filename, openconfig_acl, "openconfig_acl")
...
  File "/usr/local/lib/python3.9/dist-packages/pyangbind/lib/serialise.py", line 252, in load_json
    raise AttributeError("JSON object contained a key that" +
AttributeError: JSON object contained a key thatdid not exist (ACLSSH|RULE_20)```

So in summary it seems to me that the ```INC_INTF``` is not breaking any functionality in acl_loader.

bingwang-ms: Hi @ZhaohuiS, @nyanto , look like `INC_INTF` is a new configuration entry in CONFIG_DB. Is there any YANG model for this? If there is no YANG model support, my concern is it will break YANG validator and fail `config reload`.
As for `acl-loader` support, I think it can be addressed in another PR. The known limitation is that we can't inject such ACL rule with `INC_INTF` by `acl-loader update acl.json`. If the new configuration is not present in `acl.json`, it will not break anything.

nyanto: I think it is a valid concern to ensure the YANG model matches, thanks @bingwang-ms and @ZhaohuiS for catching this. I propose that I update this PR to use the ```IN_PORTS``` key to represent the incoming interface, in place of ```INC_INTF```, seeing that the former is already present in the YANG model (sonic-acl.yang). If you all agree that this is the right thing to do.  
Alternatively, the YANG model and related things could be changed to ```INC_INTF```, but that probably would require touching more things.

ZhaohuiS: @nyanto I am fine with this proposal, @bingwang-ms  what do you think about it?

nyanto: @ZhaohuiS  I have added to this PR a commit based on the above proposal.

nyanto: Hi @ZhaohuiS @bingwang-ms any feedback on my latest commit changing the name of the key to conform to existing YANG model?

ZhaohuiS: @nyanto Any test case update or test gap for this change?

nyanto: Yes @ZhaohuiS the key name in the related test case was updated as part of the commit.
https://github.com/sonic-net/sonic-host-services/pull/195/commits/0595b5eaec3ffa012289d90f20a4e3823683743d

ZhaohuiS: @nyanto any update in this file? https://github.com/sonic-net/sonic-mgmt/blob/master/tests/cacl/test_cacl_application.py

nyanto: Thanks @ZhaohuiS for pointing to this file, I was not aware of this set of tests. It seems to me that the the scope of the tests in this file is to validate the default ("expected") cacl. So I don't think it needs to be updated on account of the additional keys supported in this PR. Please let me know if there is a case I missed.

ZhaohuiS (approved):

madhupalu:
@anilpannala @zhangyanzhao Could you please help this PR merge?

madhupalu:
Thanks @qiluo-msft for merging the PR.

nyanto:
thanks @madhupalu and all!

PR #197, opened 2024-12-27
caclmgrd interface rules patch 1

gupurush:
Added management port exclusion (!, -i, eth0) for data plane protocols in iptable rules. Data plane protocols like BGP, BFD, and VXLAN are restricted to data interfaces only, while management protocols (NTP, SNMP, SSH) retain management port access. This enforces proper traffic segregation between management and data plane.

**What is the motivation for this PR?**
To enhance security by properly segregating management and data plane traffic through iptable rules in caclmgrd.

**How did you do it?**
Added exclude_mgmt_port rule (!, -i, eth0)
Applied exclusion to data plane protocols (BGP, BFD, VXLAN)
Preserved management port access for management services (NTP, SNMP, SSH)
Implemented conditional exclusion for other ACL rules based on service type

### DIFF
--- a/scripts/caclmgrd
+++ b/scripts/caclmgrd
@@ -791,6 +791,11 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
                             # Append the packet action as the jump target
                             rule_cmd += ["-j", "{}".format(rule_props["PACKET_ACTION"])]
 
+                            if (dst_port not in self.ACL_SERVICES["NTP"]["dst_ports"] and
+                                dst_port not in self.ACL_SERVICES["SNMP"]["dst_ports"] and
+                                dst_port not in self.ACL_SERVICES["SSH"]["dst_ports"]):
+                                rule_cmd = self.exclude_mgmt_port(rule_cmd)
+
                             iptables_cmds.append(self.iptables_cmd_ns_prefix[namespace] + rule_cmd)
                             num_ctrl_plane_acl_rules += 1
 
--- a/scripts/caclmgrd
+++ b/scripts/caclmgrd
@@ -147,6 +147,7 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
         self.namespace_mgmt_ipv6 = self.get_namespace_mgmt_ipv6(self.iptables_cmd_ns_prefix[DEFAULT_NAMESPACE], DEFAULT_NAMESPACE)
         self.namespace_docker_mgmt_ip = {}
         self.namespace_docker_mgmt_ipv6 = {}
+        self.exclude_mgmt_port_rule = [ '!', '-i', 'eth0' ]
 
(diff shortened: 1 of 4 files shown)


scripts/caclmgrd
```diff
@@ -785,6 +791,11 @@ class ControlPlaneAclManager(daemon_base.DaemonBase):
                             # Append the packet action as the jump target
                             rule_cmd += ["-j", "{}".format(rule_props["PACKET_ACTION"])]
 
+                            if (dst_port not in self.ACL_SERVICES["NTP"]["dst_ports"] and
+                                dst_port not in self.ACL_SERVICES["SNMP"]["dst_ports"] and
+                                dst_port not in self.ACL_SERVICES["SSH"]["dst_ports"]):
+                                rule_cmd = self.exclude_mgmt_port(rule_cmd)
```
ZhaohuiS, line 797: For EXTERNAL_CLIENT, there is no DST_PORT setting for it, with your change, does it mean it will block restapi or gnmi traffic from data plane?

@qiluo-msft  @prsunny could you please review this change? It will not prohibit traffic from eth0 except NTP,SNMP and SSH.


```
        "EXTERNAL_CLIENT": {
            "ip_protocols": ["tcp"],
            "multi_asic_ns_to_host_fwd":True
        },
```

gupurush: @ZhaohuiS We can add the protocols in ACL_SERVICES that has to go through eth0, could we have the defined list of protocols?

ZhaohuiS: As I know, EXTERNAL_CLIENT is designed to be used for restapi and gnmi, and the port is not defined in ACL_SERVICES , we can define flexible ports for them in acl.json.

prsunny: agree with @ZhaohuiS , this can break existing scenarios.

gupurush: Removed check so EXTERNAL_CLIENT is not affected; Updated tests. Please help review

ZhaohuiS: @gupurush could you please also update test case, otherwise submodule advance PR test will fail,

https://github.com/sonic-net/sonic-mgmt/blob/master/tests/cacl/test_cacl_application.py

The best solution should be:
1. skip test_cacl_application from PR test
2. imagebuild PR get merged
3. update test case
4. add test_cacl_application back for PR test

ZhaohuiS (approved):

scripts/caclmgrd
ZhaohuiS: @gupurush  could you please update test_cacl_application.py to support your change in this PR? And provide the test evidence here, just want to make sure the image code change could be verified and no regression will be introduced. Thanks.

ZhaohuiS: @gupurush please update the test_cacl_application and enable it for PR test.

PR #208, opened 2025-01-24
Register image_service and docker_service to sonic-hostservice

hdwhdw:
image_service and docker_service was implemented but never registered to the host-service server.


xincunli-sonic (approved):
LGTM

PR #210, opened 2025-02-05
Return proper errno for image_service.set_next_boot

hdwhdw:
Returns ENOENT (2) when detect `sonic-installer list` outputs "ERROR: Image not exist".

### DIFF
--- a/host_modules/image_service.py
+++ b/host_modules/image_service.py
@@ -178,6 +178,9 @@ class ImageService(host_service.HostModule):
         if result.returncode:
             logger.error("Failed to set next boot image: {}".format(result.stderr.decode()))
             msg = result.stderr.decode()
+            # sonic-installer might not return a proper error code, so we need to check the message.
+            if "not" in msg.lower() and ("exist" in msg.lower() or "found" in msg.lower()):
+                return errno.ENOENT, msg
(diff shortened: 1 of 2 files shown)


host_modules/image_service.py
```diff
@@ -178,6 +178,9 @@ def set_next_boot(self, image):
         if result.returncode:
             logger.error("Failed to set next boot image: {}".format(result.stderr.decode()))
             msg = result.stderr.decode()
+            # sonic-installer might not return a proper error code, so we need to check the message.
+            if "not" in msg.lower() and ("exist" in msg.lower() or "found" in msg.lower()):
```
FengPan-Frank, line 182: similar to gnmi changes, I think complete snippet string is better than such combination check, there may be more corner cases for this check.

xincunli-sonic (approved):
LGTM

host_modules/image_service.py
FengPan-Frank: I mean such partial check would have some corner case, for example the msg is "not valid image" or "invalid image is found" but seems no perfect solution for this unless structural error code defined. we may go with this first.

FengPan-Frank (approved):

PR #215, opened 2025-02-07
Fix sudo command failed because root user password expired by password hardening feature issue.

liuh-80:
Fix sudo command failed because root user password expired by password hardening feature issue.

#### Why I did it
some sudo command failed after enable password hardening feature:

admin@vlab-01:~$ sudo config reload -y
sudo: Account or password is expired, reset your password and try again
Changing password for root.
Current password:

This is because:
1. root user password change time is image build time.
2. password hardening feature will set expiration time smaller than 1 year, in sonic-mgmt test it's 100 days.
3. some SONiC command will run another commands, those sub-commands will run with root user.
4. any command run by 'root' user will be blocked and ask update password.
5. SONiC diabled the 'root' user, this user not have password, also can't login with SSH/Console, which means when ask change password by the sudo command,  there is no way to provide the 'current' password user will be blocked there.

The only solution is to run 'sudo passwd root' command, this command will not ask current password, so user can set a new root password.

##### Work item tracking
- Microsoft ADO: 31221102

#### How I did it
Remove the 'root' user expiration code.

#### How to verify it
Manually verified.
Pass all test case.

- [ ] 201811
- [ ] 201911
- [ ] 202006
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305

#### Tested branch (Please provide the tested image version)

will updated with this PR image later.
- [] SONiC.master-16482.360728-2c8b4066f

Fix hostcfgd modify_single_file method generates empty file issue

### DIFF
--- a/scripts/hostcfgd
+++ b/scripts/hostcfgd
@@ -1019,7 +1019,6 @@ class PasswHardening(object):
             if account_number >= uid_min and account_number <= uid_max:
                 normal_accounts.append(account_spl[ACCOUNT_NAME])
 
-        normal_accounts.append('root') # root is also a candidate to be age modify.
         return normal_accounts
 
     def modify_passw_conf_file(self):


liuh-80:
@davidpil2002 , can you review this PR?

davidpil2002 (approved):
I double check and yes root user have no password, so no need for expiring.

qiluo-msft (approved):

StormLiangMS:
hi @liuh-80  what's the issue if we don't port back this to 202305?

liuh-80:
> hi @liuh-80 what's the issue if we don't port back this to 202305?

The only issue is this bug does not fix on 202305.
The test case already ignored on 202305 so there will no more nightly test failure.
Also, the feature does not enable, so will no impact.

PR #225, opened 2025-02-24
Use source interface configured under radius server for sending access request

lakshmi-nexthop:
Fixes Issue: https://github.com/sonic-net/sonic-buildimage/issues/21386

When authentication is triggered by the DUT, source interface configured under radius needs to be used for sending access request.

config db needs to be looked for the source interface and store the IP address of this interface.

Testing:

- Configured AAA authentication, configured radius with source interface as uplink port channel.
- Configured Radius server route is via mgmt interface.
- initiate ssh into the dut. this triggers radius auth request
- access request is sent out with port channel ip address as source.

### DIFF
--- a/scripts/hostcfgd
+++ b/scripts/hostcfgd
@@ -678,9 +678,9 @@ class AaaCfg(object):
             "RADIUS_SERVER|{}: src_intf found. Ignoring src_ip".format(addr))
                     # RADIUS: If server.src_intf, then get the corresponding
                     # src_ip based on the server.ip, and set it.
-                    src_ip = self.get_interface_ip(server['src_intf'], addr)
-                    if len(src_ip) > 0:
-                        server['src_ip'] = src_ip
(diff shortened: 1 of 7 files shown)


scripts/hostcfgd
```diff
@@ -439,7 +440,7 @@ class AaaCfg(object):
             if new_ipv4_addr != "" and new_ipv6_addr != "":
                 break
             ip_str = it[1].split("/")[0]
-            ip_addr = ipaddress.IPAddress(ip_str)
+            ip_addr = ipaddress.ip_address(ip_str)
             # Pick the first IP address from the table that matches the source interface
```
asha-behera, line 444: We should also have a check to not use link local address like below

          # Don't use link local IP for source address
            if ip_addr.is_link_local:
                continue

lakshmi-nexthop: Thanks for the review. will send a new PR once tests are run.

PR #227, opened 2025-03-06
rust version implementation of procdockerstatsd

FengPan-Frank:


scripts/rs/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "procdockerstatsd_rs"
+version = "1.0.0"
+edition = "2024"
```
qiluo-msft, line 4: >2024 [](http://example.com/codeflow?start=11&length=4)

Should it be 2025? minor issue.

scripts/rs/main.rs
```diff
@@ -0,0 +1,212 @@
...
+use sysinfo::{System, Process};
+use chrono::Utc;
+use std::fs;
+use std::collections::HashMap;
+use std::time::SystemTime;
+use sysinfo::{SystemExt, ProcessExt};
+use std::collections::{HashSet};
+use std::time::{Duration, Instant};
+
+const REDIS_URL: &str = "redis://127.0.0.1/";
```
qiluo-msft, line 14: >127 [](http://example.com/codeflow?start=33&length=3)

magic string. May reuse rust lib for swss-common, and get the SonicDBConfig.

scripts/rs/main.rs
```diff
@@ -0,0 +1,212 @@
...
+    }
+}
+
+fn convert_to_bytes(value: &str) -> u64 {
+    let re = Regex::new(r"(\d+\.?\d*)([a-zA-Z]+)").unwrap();
+    if let Some(caps) = re.captures(value) {
+        let num: f64 = caps[1].parse().unwrap_or(0.0);
+        let unit = &caps[2];
+        match unit.to_lowercase().as_str() {
+            "kb" => (num * 1000.0) as u64,
```
qiluo-msft, line 34: >1000 [](http://example.com/codeflow?start=27&length=4)

Is it 1024?

scripts/rs/main.rs
```diff
@@ -0,0 +1,212 @@
...
+            stats_list.push(stats);
+        }
+    }
+    stats_list
+}
+
+fn collect_docker_stats(conn: &mut Connection) {
+    if let Some(output) = run_command(&["docker", "stats", "--no-stream", "-a"]) {
+        let stats_list = parse_docker_stats(&output);
+        let _: () = redis::cmd("DEL").arg("DOCKER_STATS|*").execute(conn);
```
qiluo-msft, line 75: >redis::cmd [](http://example.com/codeflow?start=20&length=10)

Can we use rust version swss-common?

scripts/rs/Cargo.toml
r12f: this is the rusttoml edition, and yea, it is 2024 for now.

scripts/rs/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "procdockerstatsd_rs"
```
r12f, line 2: procdockerstatsd-rs is more rusty : D

scripts/rs/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+version = "1.0.0"
+edition = "2024"
+description = "Daemon to collect process and Docker stats and push them to Redis"
+license = "MIT"
+authors = ["Feng Pan"]
+
+[dependencies]
+redis = "0.23"
```
r12f, line 10: we have wrapped the swss-common rust APIs here: https://github.com/sonic-net/sonic-dash-ha/tree/master/crates/swss-common, so I wonder if we can use it instead of using redis package directly.

scripts/rs/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
```
r12f, line 1: should we update the folder to procdockerstatsd to be more explicit?

if we like a single crate to have multiple bin, then updating the main.rs into procdockerstatsd.rs will be better for future.

scripts/rs/main.rs
```diff
@@ -0,0 +1,159 @@
...
+    if output.status.success() {
+        Some(String::from_utf8_lossy(&output.stdout).to_string())
+    } else {
+        eprintln!("Error running command: {:?}", cmd);
+        None
+    }
+}
+
+fn convert_to_bytes(value: &str) -> u64 {
+    let re = Regex::new(r"(\d+\.?\d*)([a-zA-Z]+)").unwrap();
```
erer1243, line 27: This recompiles the regex on every call. This is bad practice, though admittedly if it's just once every 2 minutes it won't matter. This code will cache it for you.
```rust
fn convert_to_bytes(value: &str) -> u64 {
    static RE: Regex = LazyLock::new(|| Regex::new(r"(\d+\.?\d*)([a-zA-Z]+)").unwrap());
    if let Some(caps) = RE.captures(value) {
    ...
```

scripts/rs/main.rs
```diff
@@ -0,0 +1,159 @@
...
+            stats.insert("NAME".to_string(), values[1].to_string());
+            stats.insert("CPU%".to_string(), values[2].trim_end_matches('%').to_string());
+            stats.insert("MEM_BYTES".to_string(), convert_to_bytes(values[3]).to_string());
+            stats.insert("MEM_LIMIT_BYTES".to_string(), convert_to_bytes(values[5]).to_string());
+            stats.insert("MEM%".to_string(), values[6].trim_end_matches('%').to_string());
+            stats.insert("NET_IN_BYTES".to_string(), convert_to_bytes(values[7]).to_string());
+            stats.insert("NET_OUT_BYTES".to_string(), convert_to_bytes(values[9]).to_string());
+            stats.insert("BLOCK_IN_BYTES".to_string(), convert_to_bytes(values[10]).to_string());
+            stats.insert("BLOCK_OUT_BYTES".to_string(), convert_to_bytes(values[12]).to_string());
+            stats.insert("PIDS".to_string(), values[13].to_string());
```
erer1243, line 63: this might be cleaner if the key was `&'static str`, no calls to .to_string() for keys. I also like to construct maps like `HashMap::from_iter([("CONTAINER_ID", values[0].to_string()), ("NAME", values[1].to_string()), ...]);` but both suggestions are just a matter of taste.

PR #239, opened 2025-04-11
Monitor DNS_OPTIONS for changes

bhouse-nexthop:
A new DNS_OPTIONS table has been added to allow for more than basic nameserver configuration.  We need to monitor this table for changes like we do DNS_NAMESERVER and simply reload the DNS configuration service when changes are detected.

Depends on https://github.com/sonic-net/sonic-buildimage/pull/22310

### DIFF
--- a/scripts/hostcfgd
+++ b/scripts/hostcfgd
@@ -2126,6 +2126,7 @@ class HostConfigDaemon:
         syslog_cfg = init_data.get(swsscommon.CFG_SYSLOG_CONFIG_TABLE_NAME, {})
         syslog_srv = init_data.get(swsscommon.CFG_SYSLOG_SERVER_TABLE_NAME, {})
         dns = init_data.get('DNS_NAMESERVER', {})
+        dns_options = init_data.get('DNS_OPTIONS', {})
         fips_cfg = init_data.get('FIPS', {})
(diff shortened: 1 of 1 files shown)


bhouse-nexthop:
@hdwhdw mind taking a look at this?

bhouse-nexthop:
@hdwhdw any chance you can get this merged?

bhouse-nexthop:
> @bhouse-nexthop yes. But we probably some unit test to clear the PR check first.

I honestly don't see how we could write a test case for this.  The existing DNS nameserver changes also don't have tests. All this does is restarts the service responsible for generating the `/etc/resolv.conf` when the config is detected to have changed.  Since that daemon resides within sonic-buildimage and not sonic-host-services, there's no way to validate the configuration that is generated by the changes or anything like that.

bhouse-nexthop:
> > @bhouse-nexthop yes. But we probably some unit test to clear the PR check first.
> 
> I honestly don't see how we could write a test case for this. The existing DNS nameserver changes also don't have tests. All this does is restarts the service responsible for generating the `/etc/resolv.conf` when the config is detected to have changed. Since that daemon resides within sonic-buildimage and not sonic-host-services, there's no way to validate the configuration that is generated by the changes or anything like that.

I added a super simple check, it appears to have satisfied the code coverage thing though.

PR #240, opened 2025-04-17
[featured] Changes for additional SpineRouter roles

tjchadaga:
Changes to allow dependent service restart optimization on additional SpineRouter roles


arlakshm (commented):
Do we still need this change?

PR #242, opened 2025-04-17
Implement Download and Delete For File Service

hdwhdw:
This is for implementing any gNOI service, such as System.SetPackage, File.TransferToRemote or Containerz.Deploy which have an option to support remote download. For remote download, the definition is according to the protobuf:
```
// RemoteDownload defines the details for a device to initiate a file transfer
// from or to a remote location.
message RemoteDownload {
  // The path information containing where to download the data from or to.
  // For HTTP(S), this will be the URL (i.e. foo.com/file.tbz2).
  // For SFTP and SCP, this will be the address:/path/to/file
  // (i.e. host.foo.com:/bar/baz).
  string path = 1;

  enum Protocol {
    UNKNOWN = 0;
    SFTP = 1;
    HTTP = 2;
    HTTPS = 3;
    SCP = 4;
  }
  Protocol protocol = 2;

  types.Credentials credentials = 3;

  // Optional source address used to initiate connections from the device.
  // It can be either an IPv4 address or an IPv6 address, depending on the
  // connection's destination address.
  string source_address = 4;

  // Optional source vrf used to initiate connections from the device.
  string source_vrf = 5;
}
```

See https://github.com/openconfig/gnoi/blob/main/common/common.proto#L43

The PR [SONiC gNMI PR #378](https://github.com/sonic-net/sonic-gnmi/pull/378) depends on it


host_modules/file_service.py
```diff
@@ -42,4 +45,106 @@ def get_file_stat(self, path):
...
+                return EXIT_FAILURE, f"File already exists: {local_path}"
+
+            # 2. The directory we are writing to must be world writable
+            dir_path = os.path.dirname(local_path) or "."
+            try:
+                dir_stat = os.stat(dir_path)
+            except Exception as e:
+                return EXIT_FAILURE, f"Directory not found: {dir_path} ({e})"
+            if not (dir_stat.st_mode & stat.S_IWOTH):
+                return EXIT_FAILURE, f"Directory is not world writable: {dir_path}"
```
saiarcot895, line 78: Why is this a requirement?

host_modules/file_service.py
```diff
@@ -42,4 +45,106 @@ def get_file_stat(self, path):
...
+                try:
+                    ssh.connect(hostname, username=username, password=password)
+                    sftp = ssh.open_sftp()
+                    sftp.get(remote_path, local_path)
+                    sftp.close()
+                    ssh.close()
+                    return 0, ""
+                except Exception as e:
+                    ssh.close()
+                    return 1, str(e)
```
saiarcot895, line 94: Use `EXIT_FAILURE` here?

host_modules/file_service.py
hdwhdw: The usage we have currently is for downloading temporary files and this is to prevent someone use this to attack say any sensitive root filesystem. Another example is https://github.com/sonic-net/sonic-host-services/blob/master/host_modules/image_service.py#L56

host_modules/file_service.py
hdwhdw: Thanks. Done.

tests/host_modules/file_test.py
```diff
@@ -0,0 +1,324 @@
...
+        mock_exists.return_value = True
+        mock_file_stat = mock.Mock()
+        mock_file_stat.st_mode = 0o100644  # Not world writable
+        mock_stat.return_value = mock_file_stat
+        file_service_stub = file_service.FileService(file_service.MOD_NAME)
+        path = "/some/file.txt"
+        ret, msg = file_service_stub.remove(path)
+        assert ret == 1
+        assert "File is not world writable" in msg
+        mock_remove.assert_not_called()
```
saiarcot895, line 324: Can you add a test case for downloading a file and then removing that file? I'm assuming that that is _expected_ to work, but because of umask, the file might not get written with world-writable permissions.

hdwhdw: Sorry this test file was stale. Updated the unit test so it verifies dir permission instead of file permission. But unit test cannot verify what you asked for because we simply mocked out file/dir permission checking, so os.stats will just output whatever I tell it to output. 

But we have indeed verified you can download and delete the sama (tmp) file in integration test in my gnmi PR. See the PR description in https://github.com/sonic-net/sonic-gnmi/pull/378

image

saiarcot895 (approved):

PR #251, opened 2025-05-06
Fix slow halt reboot test

jkmar:
TBD


jkmar:
Please review @kwangsuk @anand-kumar-subramanian @mbalachandar @amrutasali @sachinholla @sneelam20

PR #256, opened 2025-05-16
Modify cacl rules to allow incoming packets to dash-ha

yue-fred-gao:
### why
swbusd in dash-ha runs grpc client and server for connections within the switch and outside the switch. We need to config CACL iptables rules if dash-ha feature is enabled.

### what this PR does
if dash-ha feature is enabled, subscribe to DPU table in config_db. Configure iptables based on the swbus_port in each DPU entries (up to 8). If swbus_port is changed, remove old iptables rules and add new rules. If DPU entry is deleted, remove corresponding iptables rules.

### DIFF
--- a/scripts/caclmgrd
+++ b/scripts/caclmgrd
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/python3
 #
 # caclmgrd
 #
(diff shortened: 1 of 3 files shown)


scripts/caclmgrd
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
```
zjswhhh, line 1: Why the change?

zjswhhh (approved):

scripts/caclmgrd
yue-fred-gao: It might be by accident. Let me put it back.

zjswhhh:
Hi @qiluo-msft - please help merge.

PR #259, opened 2025-05-20
Improve reboot status tracking and add reboot count & method to status

vvolam:
# Why I did it
To provide more accurate and detailed tracking of reboot operations on the host. This includes:

- Fixing incorrect or missing values in the reboot_status_flag.
- Adding the ability to track how many times a reboot has been initiated (count).
- Recording which reboot method was used (method), which is critical for debugging and telemetry.

Fixes: https://github.com/sonic-net/sonic-buildimage/issues/22157

# How I did it
- Enhanced the populate_reboot_status_flag() function to also track:
-- count: Number of reboots initiated by the host service.
-- method: Type of reboot (e.g., cold, warm, halt).

- Corrected how container status is evaluated using Docker SDK by checking attrs['State']['Running'] instead of relying on the status string.
- Updated all relevant call sites of populate_reboot_status_flag() to pass the correct parameters.
- Adjusted unit tests to validate updated logic and ensure coverage of new parameters.

# Changelog
- Improved reboot_status_flag tracking with method and count fields.
- More accurate container status check using Docker SDK.
- Fixed bug where reboot failure didn’t update the status properly.
- Updated unit tests accordingly.

### DIFF
--- a/host_modules/reboot.py
+++ b/host_modules/reboot.py
@@ -8,6 +8,7 @@ import docker
 import psutil
 from host_modules import host_service
 from utils.run_cmd import _run_command
+from enum import Enum
 
 MOD_NAME = 'reboot'
 # Reboot method in reboot request
@@ -24,6 +25,12 @@ HALT_TIMEOUT = 60
 EXECUTE_COLD_REBOOT_COMMAND = "sudo reboot"
 EXECUTE_HALT_REBOOT_COMMAND = "sudo reboot -p"
 EXECUTE_WARM_REBOOT_COMMAND = "sudo warm-reboot"
+
+class RebootStatus(Enum):
+    STATUS_UNKNOWN = 0
+    STATUS_SUCCESS = 1
+    STATUS_RETRIABLE_FAILURE = 2
+    STATUS_FAILURE = 3
 
 logger = logging.getLogger(__name__)
 
@@ -46,7 +53,7 @@ class Reboot(host_service.HostModule):
         self.populate_reboot_status_flag()
         super(Reboot, self).__init__(mod_name)
 
-    def populate_reboot_status_flag(self, active = False, when = 0, reason = "", method = ""):
+    def populate_reboot_status_flag(self, active = False, when = 0, reason = "", method = "", status = RebootStatus.STATUS_UNKNOWN):
         """Populates the reboot_status_flag with given input params"""
         self.lock.acquire()
         self.reboot_status_flag["active"] = active
@@ -54,6 +61,7 @@ class Reboot(host_service.HostModule):
         self.reboot_status_flag["reason"] = reason
         self.reboot_status_flag["count"] = self.reboot_count
         self.reboot_status_flag["method"] = method
+        self.reboot_status_flag["status"] = status
         self.lock.release()
         return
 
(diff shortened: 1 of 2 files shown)


host_modules/reboot.py
```diff
@@ -71,10 +77,10 @@ def is_container_running(self, container_name):
         """Check if a given container is running using the Docker SDK."""
         try:
             client = docker.from_env()
-            containers = client.containers.list(filters={"name": container_name})
+            containers = client.containers.list(filters={"name": f"^{container_name}$"})
 
             for container in containers:
-                if container.name == container_name and container.status == "running":
+                if container.name == container_name and container.attrs['State']['Running']:
```
hdwhdw, line 83: Just curious, what's the difference here.

host_modules/reboot.py
```diff
@@ -111,7 +117,7 @@ def execute_reboot(self, reboot_method):
 
         rc, stdout, stderr = _run_command(command)
         if rc:
-            self.populate_reboot_status_flag()
+            self.populate_reboot_status_flag(False, int (time.time()), "Failed to execute reboot command", reboot_method)
```
hdwhdw, line 120: Nits: can we have the return code in the error message as well?

hdwhdw (approved):

host_modules/reboot.py
vvolam: Both are almost same change but later code is giving more accurate and real time results

host_modules/reboot.py
vvolam: I have updated the reboot_status with reboot method status. Is that what you are expecting?

host_modules/reboot.py
```diff
@@ -38,16 +45,23 @@ def __init__(self, mod_name):
         self.lock = threading.Lock()
         # reboot_status_flag is used to keep track of reboot status on host
         self.reboot_status_flag = {}
+
+        # reboot count
+        self.reboot_count = 0
```
kishanps, line 50: There is a reboot_count tracked in rebootbe, why an additional count here ?

vvolam: To populate this information in reboot_status_flag()

host_modules/reboot.py
```diff
@@ -128,29 +142,30 @@ def execute_reboot(self, reboot_method):
             logger.info("%s: Waiting until services are halted or timeout occurs", MOD_NAME)
             timeout = HALT_TIMEOUT
             start_time = time.monotonic()
+
             while time.monotonic() - start_time < timeout:
                 if not self.is_halt_command_running() and not self.is_container_running("pmon"):
                     logger.info("%s: Halting the services is completed on the device", MOD_NAME)
+                    self.populate_reboot_status_flag(False, 0, "Halt reboot completed", reboot_method, RebootStatus.STATUS_SUCCESS)
```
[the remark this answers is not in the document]
vvolam: It is blocking call, but pmon process is running few more secs after the script execution. so I had to keep this loop.

tests/host_modules/reboot_test.py
```diff
@@ -37,6 +39,12 @@
 VALID_REBOOT_REQUEST_WARM = "{\"method\": \"WARM\", \"message\": \"test reboot request reason\"}"
 INVALID_REBOOT_REQUEST = "\"method\": 1, \"message\": \"test reboot request reason\""
 
+class RebootStatus(Enum):
```
qiluo-msft, line 42: >RebootStatus [](http://example.com/codeflow?start=6&length=12)

Could you reuse this enum?

PR #228, opened 2025-03-11
Pin PyGObject to 3.50.0

cpackham-atlnz:
Newer versions of PyGObject depend on libgirepository-2.0-dev which isn't packaged for Debian. This leads to build failures like

  Collecting PyGObject
  Downloading pygobject-3.52.1.tar.gz (1.2 MB)
  ...

  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [24 lines of output]
      ...

      ../meson.build:31:9: ERROR: Dependency 'girepository-2.0' is required but not found.

Pin the version of PyGObject to 3.50.0 which is the latest version known to work.

Fixes sonic-net/sonic-buildimage#21984

PR #264, opened 2025-06-02
Fix PyGObject required version

qiluo-msft:
Current slave image installs an old version.
files/build/versions/dockers/sonic-slave-bookworm/versions-py3:113:pygobject==3.42.2

Relax the install_requires range of PyGObject to allow any versions.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,7 @@ setup(
         'dbus-python',
         'systemd-python',
         'Jinja2>=2.10',
-        'PyGObject==3.50.0',
+        'PyGObject',
         'pycairo==1.26.1',
         'psutil'
     ] + sonic_dependencies,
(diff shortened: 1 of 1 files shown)


saiarcot895:
Can we get rid of the version range requirement itself? I'm guessing that the code in this package wouldn't break if a version of `pygobject` newer than 3.50.0 were installed.

qiluo-msft:
> Can we get rid of the version range requirement itself? I'm guessing that the code in this package wouldn't break if a version of `pygobject` newer than 3.50.0 were installed.

Reading https://github.com/sonic-net/sonic-host-services/pull/228, seems there is a good reason to limit <= 3.50.0

saiarcot895 (approved):

PR #270, opened 2025-06-11
Add gNOI OS Management services

Kanchana-HCLTech:


kanchanavelusamy:
> Can you see if you can take advantage of https://github.com/sonic-net/sonic-host-services/blob/master/host_modules/image_service.py? We should avoid writing different modules for the same service. Thanks!

Hi @hdwhdw,
Thanks for the suggestion.
You're right that we should avoid duplicating logic across services. I looked closely at both implementations, and here’s what I found:
The install() method in image_service.py is a simple D-Bus method that wraps a single call to sonic-installer install, accepting a local file path or URL.
In contrast, gnoi_os_mgmt.py implements a multi-phase install() method based on the gNOI os.proto Install RPC, which involves:
Parsing structured JSON input
Handling TransferRequest, TransferContent, and TransferEnd steps
Supporting chunked transfers (in FE/BE model)
Managing intermediate state (version_in_progress, locks)

So while both modules have a method named install(), they serve different protocol and interface layers.

kanchanavelusamy:
> Please take a look at existing services to see if you can reuse. We should avoid reinventing the wheel. If the implementation here is much more robust feel free to propose substituting the existing service (and also in sonic-gnmi), but we should avoid having two services doing the same thing.

As explained, the gNOI implementation follows a more protocol-compliant approach and isn't a duplicate of image_service.py

kanchanavelusamy:
@ndas7: Could you please review and share your feedback?

cc:@kishanps

kanchanavelusamy:
> LGTM but can you update your PR description to reflect what is actually in the PR? Thanks.

Addressed. Thanks Dawei.

ndas7 (approved):

xumia:
@kanchanavelusamy , when to support OS installation feature?

PR #234, opened 2025-03-31
[featured] fix non existing feature start

stepanblyschak:
Fix an issue that when starting a non existing feature and then installing it with sonic-package-manager it won't start.

E.g:

```
admin@sonic:~$ redis-cli -n 4 hmset 'FEATURE|non-existing' state enabled OK
admin@sonic:~$ sudo zless /var/log/syslog | grep featured 2025 Mar 26 15:19:33.237090 sonic INFO featured: Feature non-existing is enabled and started
```

With this patch an error is recordded and feature state is not cached:

```
2025 Mar 26 15:35:32.794640 sonic INFO featured: Running cmd: '['sudo', 'systemctl', 'unmask', 'non-existing.service']'
2025 Mar 26 15:35:33.233685 sonic INFO featured: Output:  , Stderr: Unit non-existing.service does not exist, proceeding anyway.
2025 Mar 26 15:35:33.233743 sonic INFO featured: Running cmd: '['sudo', 'systemctl', 'enable', 'non-existing.service']'
2025 Mar 26 15:35:33.247238 sonic ERR featured: ['sudo', 'systemctl', 'enable', 'non-existing.service'] - failed: return code - 1, output:
2025 Mar 26 15:35:33.247300 sonic ERR featured: Feature 'non-existing.service' failed to be enabled and started
2025 Mar 26 15:35:39.331762 sonic INFO featured: Running cmd: '['sudo', 'systemctl', 'unmask', 'non-existing.service']'
2025 Mar 26 15:35:39.780474 sonic INFO featured: Output:  , Stderr: Unit non-existing.service does not exist, proceeding anyway.
2025 Mar 26 15:35:39.780513 sonic INFO featured: Running cmd: '['sudo', 'systemctl', 'enable', 'non-existing.service']'
2025 Mar 26 15:35:39.795113 sonic ERR featured: ['sudo', 'systemctl', 'enable', 'non-existing.service'] - failed: return code - 1, output:
2025 Mar 26 15:35:39.795153 sonic ERR featured: Feature 'non-existing.service' failed to be enabled and started
```

Feature will start next time when setting state field in FEATURE table

PR #272, opened 2025-06-20
Revert "[featured] fix non existing feature start (#234)"

qiluo-msft:
This reverts commit 29b8be81fd4f468ff8482df37b372dafe3c07b3a.

Because it blocks submodule update in sonic-buildimage repo.
The build and test are verified in a draft PR https://github.com/sonic-net/sonic-buildimage/pull/23027.


qiluo-msft:
This is a clean reverting PR, so the coverage check should not block. I will bypapss coverage check and merge

PR #274, opened 2025-06-25
Integrated Security Cipher module with Hostcfgd for passkey decryption

nmoray:
- What I did
Made use of Security Cipher module to decrypt the encrypted TACPLUS passkey

- How I did it
Implemented the feature by following [HLD](https://github.com/sonic-net/SONiC/pull/1471)

- How to verify it
```
1. Encrypt passkey:
sonic# config tacacs passkey nikhil --encrypt
sonic# show run al | jq .TACPLUS
{
  "global": {
    "auth_type": "login",
    "key_encrypt": "True",
    "passkey": "U2FsdGVkX19YTUcG+0r3+d78A5E0zTXEukS3ZNrfyFk=",
    "src_intf": "Loopback0"
  }
}
sonic# cat /etc/pam.d/common-auth-sonic
#THIS IS AN AUTO-GENERATED FILE
#
# /etc/pam.d/common-auth- authentication settings common to all services
# This file is included from other service-specific PAM config files,
# and should contain a list of the authentication modules that define
# the central authentication scheme for use on the system
# (e.g., /etc/shadow, LDAP, Kerberos, etc.). The default is to use the
# traditional Unix authentication mechanisms.
#
# here are the per-package modules (the "Primary" block)

auth	[success=done new_authtok_reqd=done default=ignore auth_err=die]	pam_tacplus.so server=<> secret=nikhil login=login timeout=5   try_first_pass
auth	[success=done new_authtok_reqd=done default=ignore auth_err=die]	pam_tacplus.so server=<> secret=nikhil login=login timeout=5   try_first_pass
auth	[success=1 default=ignore]	pam_unix.so nullok try_first_pass

#
# here's the fallback if no module succeeds
auth    requisite                       pam_deny.so
# prime the stack with a positive return value if there isn't one already;
# this avoids us returning an error just because nothing sets a success code
# since the modules above will each just jump around
auth    required                        pam_permit.so
# and here are more per-package modules (the "Additional" block)
sonic# cat /etc/cipher_pass.json
{
  "TACPLUS": {
    "table_info": [
      "TACPLUS|global"
    ],
    "password": "TEST1"
  }

2. Encrypt passkey and update existing password
sonic# config tacacs passkey nikhil --encrypt --rotate
Password:
sonic# cat /etc/cipher_pass.json
{
  "TACPLUS": {
    "table_info": [
      "TACPLUS|global"
    ],
    "password": "TEST3"
  }
sonic# cat /etc/pam.d/common-auth-sonic
#THIS IS AN AUTO-GENERATED FILE
#
# /etc/pam.d/common-auth- authentication settings common to all services
# This file is included from other service-specific PAM config files,
# and should contain a list of the authentication modules that define
# the central authentication scheme for use on the system
# (e.g., /etc/shadow, LDAP, Kerberos, etc.). The default is to use the
# traditional Unix authentication mechanisms.
#
# here are the per-package modules (the "Primary" block)

auth	[success=done new_authtok_reqd=done default=ignore auth_err=die]	pam_tacplus.so server=<> secret=nikhil login=login timeout=5   try_first_pass
auth	[success=done new_authtok_reqd=done default=ignore auth_err=die]	pam_tacplus.so server=<> secret=nikhil login=login timeout=5   try_first_pass
auth	[success=1 default=ignore]	pam_unix.so nullok try_first_pass

#
# here's the fallback if no module succeeds
auth    requisite                       pam_deny.so
# prime the stack with a positive return value if there isn't one already;
# this avoids us returning an error just because nothing sets a success code
# since the modules above will each just jump around
auth    required                        pam_permit.so
# and here are more per-package modules (the "Additional" block)
sonic# show run al | jq .TACPLUS
{
  "global": {
    "auth_type": "login",
    "key_encrypt": "True",
    "passkey": "U2FsdGVkX18/8nFrAiCe01pOqgPoUUZFaTpmtawdi8I=",
    "src_intf": "Loopback0"
  }
}

3. Revert back to plaintext passkey
sonic# config tacacs passkey ravi
sonic# cat /etc/pam.d/common-auth-sonic
#THIS IS AN AUTO-GENERATED FILE
#
# /etc/pam.d/common-auth- authentication settings common to all services
# This file is included from other service-specific PAM config files,
# and should contain a list of the authentication modules that define
# the central authentication scheme for use on the system
# (e.g., /etc/shadow, LDAP, Kerberos, etc.). The default is to use the
# traditional Unix authentication mechanisms.
#
# here are the per-package modules (the "Primary" block)

auth	[success=done new_authtok_reqd=done default=ignore auth_err=die]	pam_tacplus.so server=<> secret=ravi login=login timeout=5   try_first_pass
auth	[success=done new_authtok_reqd=done default=ignore auth_err=die]	pam_tacplus.so server=<> secret=ravi login=login timeout=5   try_first_pass
auth	[success=1 default=ignore]	pam_unix.so nullok try_first_pass

#
# here's the fallback if no module succeeds
auth    requisite                       pam_deny.so
# prime the stack with a positive return value if there isn't one already;
# this avoids us returning an error just because nothing sets a success code
# since the modules above will each just jump around
auth    required                        pam_permit.so
# and here are more per-package modules (the "Additional" block)
sonic# show run al | jq .TACPLUS
{
  "global": {
    "auth_type": "login",
    "key_encrypt": "False",
    "passkey": "ravi",
    "src_intf": "Loopback0"
  }
}
sonic# cat /etc/cipher_pass.json
{
  "TACPLUS": {
    "table_info": [],
    "password": null
  }
```

- Related PR(s)
- [PR#17201](https://github.com/sonic-net/sonic-buildimage/pull/17201)
- [PR#22711](https://github.com/sonic-net/sonic-buildimage/pull/22711)
- [PR#3936](https://github.com/sonic-net/sonic-utilities/pull/3936)

Note:: This PR is created by closing the old [PR](https://github.com/sonic-net/sonic-host-services/pull/81)

### DIFF
--- a/scripts/hostcfgd
+++ b/scripts/hostcfgd
@@ -23,6 +23,7 @@ hostcfg_file_path = os.path.abspath(__file__)
 hostcfg_dir_path = os.path.dirname(hostcfg_file_path)
 sys.path.append(hostcfg_dir_path)
 import ldap
+from sonic_py_common.security_cipher import master_key_mgr 
 
 # FILE
 PAM_AUTH_CONF = "/etc/pam.d/common-auth-sonic"
@@ -96,7 +97,6 @@ DEFAULT_FIPS_RESTART_SERVICES = ['ssh', 'telemetry.service', 'restapi']
 # MISC Constants
 CFG_DB = "CONFIG_DB"
 STATE_DB = "STATE_DB"
-
 
 def signal_handler(sig, frame):
     if sig == signal.SIGHUP:
@@ -647,14 +647,20 @@ class AaaCfg(object):
             src_ip = None
 
         servers_conf = []
+        secure_cipher = master_key_mgr()
         if self.tacplus_servers:
             for addr in self.tacplus_servers:
                 server = tacplus_global.copy()
                 server['ip'] = addr
                 server.update(self.tacplus_servers[addr])
+                if server['key_encrypt'] == 'True':
+                    output = secure_cipher.decrypt_passkey("TACPLUS", server['passkey'])
+                    if output:
+                       server['passkey'] = output
+                    else:
+                        syslog.syslog(syslog.LOG_ERR, "Decrypt_passkey failed for TACPLUS Server: {}.".format(addr))
                 servers_conf.append(server)
             servers_conf = sorted(servers_conf, key=lambda t: int(t['priority']), reverse=True)
-
         radius_global = self.radius_global_default.copy()
         radius_global.update(self.radius_global)
 
--- a/scripts/hostcfgd
+++ b/scripts/hostcfgd
@@ -653,7 +653,7 @@ class AaaCfg(object):
                 server = tacplus_global.copy()
                 server['ip'] = addr
                 server.update(self.tacplus_servers[addr])
-                if server['key_encrypt'] == 'True':
+                if 'key_encrypt' in server and server['key_encrypt'] == 'True':
                     output = secure_cipher.decrypt_passkey("TACPLUS", server['passkey'])
                     if output:
                        server['passkey'] = output


nmoray:
@anders-nexthop please help in adding the reviewers.

scripts/hostcfgd
```diff
@@ -647,14 +647,20 @@ class AaaCfg(object):
...
 
         servers_conf = []
+        secure_cipher = master_key_mgr()
         if self.tacplus_servers:
             for addr in self.tacplus_servers:
                 server = tacplus_global.copy()
                 server['ip'] = addr
                 server.update(self.tacplus_servers[addr])
+                if 'key_encrypt' in server and server['key_encrypt'] == 'True':
+                    output = secure_cipher.decrypt_passkey("TACPLUS", server['passkey'])
```
anders-nexthop, line 657: shouldnt this be TACPLUS_SERVER?

anders-nexthop (commented):
One comment.

I am not very familiar with this code, but I'm wondering if this is the right place to decrypt the passkey? isn't this just going to generate more config files? would it be better to patch the actual TACACS server code to decrypt the passkey, so we aren't _ever_ storing it in plaintext in config files?

anders-nexthop:
@qiluo-msft please take a look at this review

scripts/hostcfgd
qiluo-msft, line 657: Crash: `decrypt_passkey` raises `ValueError` when no password is set in `cipher_pass.json` (e.g. after a fresh boot before the cipher is initialized). This is not caught, so hostcfgd will crash. Wrap in try/except.

scripts/hostcfgd
```diff
@@ -647,14 +647,20 @@ class AaaCfg(object):
...
+                    if output:
+                       server['passkey'] = output
```
qiluo-msft, line 659: Indentation inconsistency: this line has 7 spaces instead of 8. Python will raise `IndentationError` at runtime.

qiluo-msft (commented):
Please add unit tests — coverage checker will block this PR without them.

qiluo-msft:
Please resolve conflicts.

PR #275, opened 2025-06-25
Improve reboot status tracking and add reboot count & method to status

vvolam:
Why I did it
To provide more accurate and detailed tracking of reboot operations on the host. This includes:

Fixing incorrect or missing values in the reboot_status_flag. Adding the ability to track how many times a reboot has been initiated (count). Recording which reboot method was used (method), which is critical for debugging and telemetry. Fixes: sonic-net/sonic-buildimage#22157

How I did it
Enhanced the populate_reboot_status_flag() function to also track: -- count: Number of reboots initiated by the host service. -- method: Type of reboot (e.g., cold, warm, halt).

Corrected how container status is evaluated using Docker SDK by checking attrs['State']['Running'] instead of relying on the status string.

Updated all relevant call sites of populate_reboot_status_flag() to pass the correct parameters.

Adjusted unit tests to validate updated logic and ensure coverage of new parameters.

### DIFF
--- a/host_modules/reboot.py
+++ b/host_modules/reboot.py
@@ -8,6 +8,7 @@ import docker
 import psutil
 from host_modules import host_service
 from utils.run_cmd import _run_command
+from enum import Enum
 
 MOD_NAME = 'reboot'
 # Reboot method in reboot request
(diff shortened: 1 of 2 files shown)


vvolam:
cherry-pick of PR https://github.com/sonic-net/sonic-host-services/pull/259

yejianquan (approved):
LGTM

PR #278, opened 2025-06-30
Include info level logs in reboot cause services

roy-nexthop:
The `process-reboot-cause` and `determine-reboot-cause` services are filtering out INFO level logs. This is unintentional according to the comment "Configure logger to log all messages INFO level and higher".

This PR calls `set_min_log_priority` with **logging.INFO (20)** instead of **sonic_logger.DEFAULT_LOG_LEVEL (21)** as to not filter out the INFO logs.

### DIFF
--- a/scripts/determine-reboot-cause
+++ b/scripts/determine-reboot-cause
@@ -9,6 +9,7 @@
 try:
     import datetime
     import json
+    import logging
     import os
     import pwd
     import re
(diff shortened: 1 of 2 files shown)


liamkearney-msft (approved):
lgtm - these looks like useful logs to have. thanks

anamehra:
Hi @abdosi , @cyw233 , its a good have candidate for 202405.

PR #280, opened 2025-07-01
Fix crash for possible race between process quit and cpu_percent() call

FengPan-Frank:
What I did
Fix crash for possible race between process quit and cpu_percent() call

Why I did it
For https://github.com/sonic-net/sonic-buildimage/issues/22746, there's 

original code is
sorted_processes = sorted(processes_all, key=lambda x: x.cpu_percent(), reverse=True)

Here, x is a psutil.Process object, and calling x.cpu_percent() internally tries to read from /proc/<pid>/stat. If the process has already exited (which happens frequently in dynamic systems), this file no longer exists, and psutil raises NoSuchProcess.

Even though there's catching exceptions in the next loop (when building the final processdata), it doesn't catch exceptions during the sort, which leads to the crash.

How I verified it
Just added try catch for possible process race issue and did unit test.


make1980 (approved):
:shipit:

PR #290, opened 2025-07-11
[202411] Fix 202411 pipeline

zbud-msft:
Cherrypicks https://github.com/sonic-net/sonic-host-services/pull/260 and https://github.com/sonic-net/sonic-host-services/pull/193

They are clean cherry picks but we need both changes in 1 PR, because without 260 cherry pick, the cherry pick of 193 will never run, and without the cherry pick of 193, the cherry pick of 260 will always fail. So there is a cyclic dependency which can be fixed by including both changes in this PR


kperumalbfn:
LGTM

PR #292, opened 2025-07-18
tests: Ensure hostcfgd tests do not access /proc/cmdline

kkosciuszkiewicz-arista:
Some hostcfgd tests override contents of /proc/cmdline using various methods. In the remaining cases, actual /proc/cmdline from the host running the tests is opened and processed.

For example, if `crashkernel=` is present on the test machine, a number of tests will try accessing config db table `KDUMP` and fail.

To avoid influencing test results by the test environment, a new fixture is added to ensure reading /proc/cmdline returns an empty line by default.


jusherma:
We're seeing this issue too and this fix works for us. Can this be merged?

PR #296, opened 2025-07-30
Modify reboot cause logic to prefer SW cause over HW cause on Kernel Panic

ymd-arista:
Currently if there's a Kernel Panic and system reboots, a random HW cause can take precedence over and mask the actual cause of reboot, which is kernel panic.

To overcome such scenarios, if SW cause is 'Kernel Panic', give it preference over HW cause.


mlok-nokia (approved):

judyjoseph (approved):
LGTM

PR #301, opened 2025-08-13
[Smartswitch] Update control plane ACL for smartswitch

gpunathilell:
The smartswitch CHASSIS_STATE_DB is hosted on the redis-chassis IP address, which needs to still accept traffic on addition of control plane ACLs (DBs should be accessible at any point of time) this ip is also needed for DPU access as well. So we have a check for smartswitch to allow traffic to the bridge-midplane IP address

##Tests
Manual tests to confirm that addition of control plane ACLs does not render the CHASSIS_STATE_DB inaccesible

### DIFF
--- a/scripts/caclmgrd
+++ b/scripts/caclmgrd
@@ -337,6 +337,10 @@ class ControlPlaneAclManager(logger.Logger):
             allow_internal_chassis_midplane_traffic.append(['iptables', '-A', 'INPUT', '-s', chassis_midplane_ip, '-d', chassis_midplane_ip, '-j', 'ACCEPT'])
             allow_internal_chassis_midplane_traffic.append(['iptables', '-A', 'INPUT', '-i', chassis_midplane_dev_name, '-j', 'ACCEPT'])
 
+        if device_info.is_smartswitch():
+            # Allow traffic to the chassis midplane IP
+            allow_internal_chassis_midplane_traffic.append(['iptables', '-A', 'INPUT', '-d', '169.254.200.254','-j', 'ACCEPT'])
+
         return allow_internal_chassis_midplane_traffic
 
     def generate_allow_internal_docker_ip_traffic_commands(self, namespace):
--- a/tests/caclmgrd/caclmgrd_chassis_midplane_test.py
+++ b/tests/caclmgrd/caclmgrd_chassis_midplane_test.py
@@ -34,17 +34,18 @@ class TestCaclmgrdChassisMidplane(TestCase):
         if not os.path.exists(DBCONFIG_PATH):
             fs.create_file(DBCONFIG_PATH) # fake database_config.json
 
-        with mock.patch("sonic_py_common.device_info.is_chassis", return_value=True):
-            with mock.patch("sonic_py_common.device_info.is_smartswitch", return_value=False):
+        mock_is_chassis = mock.MagicMock(return_value=True)
+        mock_is_smartswitch = mock.MagicMock(return_value=False)
+
+        with mock.patch("sonic_py_common.device_info.is_chassis", mock_is_chassis):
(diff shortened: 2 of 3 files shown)


oleksandrivantsiv (approved):

scripts/caclmgrd
```diff
@@ -337,6 +337,10 @@ class ControlPlaneAclManager(logger.Logger):
             allow_internal_chassis_midplane_traffic.append(['iptables', '-A', 'INPUT', '-s', chassis_midplane_ip, '-d', chassis_midplane_ip, '-j', 'ACCEPT'])
             allow_internal_chassis_midplane_traffic.append(['iptables', '-A', 'INPUT', '-i', chassis_midplane_dev_name, '-j', 'ACCEPT'])
 
+        if device_info.is_smartswitch():
+            # Allow traffic to the chassis midplane IP
+            allow_internal_chassis_midplane_traffic.append(['iptables', '-A', 'INPUT', '-d', '169.254.200.254','-j', 'ACCEPT'])
```
qiluo-msft, line 342: >169.254.200.254 [](http://example.com/codeflow?start=94&length=15)

Can we get the IP from ConfigDB.
The path is "MID_PLANE_BRIDGE"/"GLOBAL"/"ip_prefix"

gpunathilell: This is a hard-coded IP Address which will never change (since it should belong in the same network as the DPU midplane ip, these are hardcoded as well)

gpunathilell: @qiluo-msft please confirm, if this is okay as the IP is hard-coded, otherwise I can parse from CONFIG_DB

qiluo-msft: If it is already in CONFIG_DB, then get from there.

scripts/caclmgrd
```diff
@@ -337,6 +361,11 @@ class ControlPlaneAclManager(logger.Logger):
             allow_internal_chassis_midplane_traffic.append(['iptables', '-A', 'INPUT', '-s', chassis_midplane_ip, '-d', chassis_midplane_ip, '-j', 'ACCEPT'])
             allow_internal_chassis_midplane_traffic.append(['iptables', '-A', 'INPUT', '-i', chassis_midplane_dev_name, '-j', 'ACCEPT'])
 
+        if device_info.is_smartswitch():
```
rameshraghupathy, line 364: @gpunathilell Do you need a namespace check for multi ASIC case?

scripts/caclmgrd
```diff
@@ -328,6 +329,29 @@ class ControlPlaneAclManager(logger.Logger):
...
+        """
+        try:
+            config_db_connector = self.config_db_map[namespace]
+            midplane_bridge_table = config_db_connector.get_table(self.MID_PLANE_BRIDGE_TABLE)
+            if midplane_bridge_table and "GLOBAL" in midplane_bridge_table:
+                global_config = midplane_bridge_table["GLOBAL"]
+                if "ip_prefix" in global_config:
+                    self.log_info("Retrieved midplane bridge IP prefix from ConfigDB: {}".format(global_config["ip_prefix"]))
+                    return global_config["ip_prefix"].split("/")[0]
+        except Exception as e:
```
qiluo-msft, line 351: >Exception [](http://example.com/codeflow?start=15&length=9)

Could you use more specific exception type?

scripts/caclmgrd
gpunathilell: No, as this is only valid for smartswitch, where we don't have multi asic funcationality

scripts/caclmgrd
gpunathilell: specified exception as runtimeerror, if there are any connectivity issues and raised exception

qiluo-msft (approved):

PR #319, opened 2025-11-03
 Fix Kdump SSH Remote Configuration Order

### LINKED ISSUE #318: Kdump SSH Remote Configuration Order Bug
## Summary
The `sonic-kdump-config --remote` command is being called before `ssh_path` and `ssh_string` are properly configured, causing the remote kdump feature to behave incorrectly.

## Problem Description
In the current implementation of the `KdumpCfg` class in `scripts/hostcfgd`, the configuration commands are executed in the wrong order:

1. **`--remote` is called first** - This command comments/uncomments the `ssh_path` and `ssh_string` configuration based on the boolean value in the CONFIG_DB KDUMP table
2. **`--ssh_string` and `--ssh_path` are called after** - These commands set the actual SSH configuration values

## Additional Issues

1. **`ssh_string` was never being set** - The code checked if `data.get("ssh_string")` exists but never assigned it to the `ssh_string` variable before calling the command
2. **`ssh_path` was never being set** - Same issue as `ssh_string`; the value was checked but never assigned
3. **`--remote` parameter missing** - The command was called with a parameter value, but should be called without parameters (the remote state is determined from CONFIG_DB)

## Expected Behavior
The configuration commands should be executed in this order:
1. Set `ssh_string` configuration
2. Set `ssh_path` configuration  
3. Call `--remote` to enable/disable remote kdump based on CONFIG_DB (which will now correctly comment/uncomment the already-configured SSH settings)

## Impact
- Remote kdump feature fails to work correctly when configured via CONFIG_DB
- SSH-based remote crash dump collection may not be enabled even when configured

arista-hpandya:
## Description
This PR fixes the ordering issue in kdump remote configuration where `sonic-kdump-config --remote` was being called before `ssh_path` and `ssh_string` were properly set, causing the remote kdump feature to behave incorrectly.

## Related Issue
Fixes #318

## Changes Made
- Reordered configuration commands to set `ssh_string` and `ssh_path` before calling `--remote`
- Fixed variable assignment for `ssh_string` - now properly assigns the value from CONFIG_DB before using it
- Fixed variable assignment for `ssh_path` - now properly assigns the value from CONFIG_DB before using it
- Removed incorrect parameter from `sonic-kdump-config --remote` call (remote state is determined from CONFIG_DB)

## Root Cause
The `sonic-kdump-config --remote` command comments/uncomments `ssh_path` and `ssh_string` in the kdump configuration file based on the remote setting in CONFIG_DB. When called before these values were set, it operated on stale or default values instead of the intended configuration.

## Testing Done
- [x] Verified kdump remote configuration via CONFIG_DB
- [x] Confirmed `ssh_string` and `ssh_path` are properly set before `--remote` is called
- [x] Tested enabling remote kdump feature
- [x] Tested disabling remote kdump feature
- [x] Verified configuration persists across service restarts

## Impact
- **Severity**: Medium
- **Component**: Kdump remote configuration
- **Affected Feature**: SSH-based remote crash dump collection

### DIFF
--- a/scripts/hostcfgd
+++ b/scripts/hostcfgd
@@ -1252,19 +1252,20 @@ class KdumpCfg(object):
                 num_dumps = data.get("num_dumps")
             run_cmd(["sonic-kdump-config", "--num_dumps", num_dumps])
 
-            # Remote option
-            remote = self.kdump_defaults["remote"]
-            if data.get("remote") is not None:
-                remote = data.get("remote")
-            run_cmd(["sonic-kdump-config", "--remote", remote])
-
+            # ssh_string
             ssh_string = self.kdump_defaults["ssh_string"]
             if data.get("ssh_string") is not None:
-                    run_cmd(["sonic-kdump-config", "--ssh_string", ssh_string])
+                ssh_string = data.get("ssh_string")
+            run_cmd(["sonic-kdump-config", "--ssh_string", ssh_string])
+
             # ssh_path
-            ssh_path= self.kdump_defaults["ssh_path"]
+            ssh_path = self.kdump_defaults["ssh_path"]
             if data.get("ssh_path") is not None:
-                    run_cmd(["sonic-kdump-config", "--ssh_path", ssh_path])
+                ssh_path = data.get("ssh_path")
+            run_cmd(["sonic-kdump-config", "--ssh_path", ssh_path])
+
+            # Remote option
+            run_cmd(["sonic-kdump-config", "--remote"])
 
 class NtpCfg(object):
     """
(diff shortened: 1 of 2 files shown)


bmridul (approved):
Pls add UT logs.

mlok-nokia (approved):

scripts/hostcfgd
```diff
@@ -1252,19 +1252,20 @@ class KdumpCfg(object):
...
+
             # ssh_path
-            ssh_path= self.kdump_defaults["ssh_path"]
+            ssh_path = self.kdump_defaults["ssh_path"]
             if data.get("ssh_path") is not None:
-                    run_cmd(["sonic-kdump-config", "--ssh_path", ssh_path])
+                ssh_path = data.get("ssh_path")
+            run_cmd(["sonic-kdump-config", "--ssh_path", ssh_path])
+
+            # Remote option
```
judyjoseph: shouldn't we get the remote option as in old code 
```
            # Remote option
            remote = self.kdump_defaults["remote"]
            if data.get("remote") is not None:
                remote = data.get("remote")
```

judyjoseph (approved):
LGTM

arlakshm:
@arista-hpandya, can you please create PR for 202511 branch

PR #343, opened 2026-01-20
[SmartSwitch] [gnoi-shutdown] Handle non-SmartSwitch platforms gracefully and simplify service dependencies

vvolam:
#### Description

This PR improves the gnoi-shutdown service by making it handle non-SmartSwitch platforms gracefully and simplifies the service startup dependencies.

#### Changes

- **Moved platform check into the daemon itself**: Instead of using `ExecCondition` with a separate `check_platform.py` script, the platform detection logic is now integrated into `gnoi_shutdown_daemon.py`. This allows the daemon to exit gracefully with a log message on non-SmartSwitch or DPU platforms, rather than failing at the systemd level.

- **Simplified service dependencies**:
  - Removed `ExecStartPre` script (`wait-for-sonic-core.sh`) that polled for service readiness
  - Added explicit `Requires` and `After` dependencies on `gnmi.service` and `pmon.service`

- **Removed unnecessary files**:
  - Deleted `check_platform.py` and its tests (`check_platform_test.py`) — logic moved into main daemon
  - Deleted `wait-for-sonic-core.sh` — replaced by proper systemd dependencies
  - Removed deleted script references from `setup.py`

- **Code cleanup**:
  - Removed unused `redis` import
  - Removed unused `STATE_DB_INDEX` constant

#### Why I did it

- The previous approach using `ExecCondition` with a separate script caused the service to fail on non-SmartSwitch platforms, which could cause noise in logs and monitoring
- The shell script for waiting on services was fragile and duplicated what systemd can handle natively through service dependencies
- Consolidating platform checks into the daemon provides better logging and cleaner service behavior

#### How Has This Been Tested?

- Unit tests added for `main()` early-exit paths: non-SmartSwitch, DPU, and platform check exception
- Existing `main_loop_flow`, `main_loop_no_dpu_name`, and `main_loop_get_transition_exception` tests updated to include `is_smartswitch`/`is_dpu` mocks
- Verified on SmartSwitch NPU platform: daemon starts and monitors CHASSIS_MODULE changes
- Verified on non-SmartSwitch platform: daemon exits gracefully with log message


saiarcot895 (approved):

gpunathilell (approved):

gpunathilell:
@vvolam please test it once to make sure there are no degradations

vvolam:
@gpunathilell Since the PR is kind of old, I have retested on smartswitch. Everything looks good.

image

PR #351, opened 2026-03-06
Fix kdump unit tests by correctly mocking `/proc/cmdline`

### LINKED ISSUE #350: Kdump unit test failures.
The following tests do no mock /proc/cmdline but rather end up reading the actual /proc/cmdline from the host erroneously- 
 
1. hostcfgd_fips_test.py
2. hostcfgd_logging_test.py
3. hostcfgd_rsyslog_test.py
4. hostcfgd_test.py
 
Even though a [cmdline](https://github.com/sonic-net/sonic-host-services/blob/master/tests/proc/cmdline) file is committed it was never used by the tests other than `hostcfgd_test.py`, instead the code ends up using the `/proc/cmdline` of the machine on which the test runs. The tests fail if the KDUMP is enabled on the machine where test is running, with a keyerror which is a false positive. As the test does not expect kdump to be enabled, the mocked config db for these tests does not have KDUMP configured as a result hostcfgd fails to retrieve KDUMP from config_db after it sees crashkernel parameteres in /proc/cmdline at this point https://github.com/sonic-net/sonic-host-services/blob/master/scripts/hostcfgd#L1195
 
The proposed fix is to mock /proc/cmdline for every test instead of reading the "real" /proc/cmdline.

```
>       host_config_daemon = hostcfgd.HostConfigDaemon()                                                                               
                                                                                                                                       
tests/hostcfgd/hostcfgd_fips_test.py:98:                                                                                               
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
scripts/hostcfgd:2171: in __init__                                                                                                     
    self.kdumpCfg = KdumpCfg(self.config_db)
scripts/hostcfgd:1177: in __init__
    self.init_kdump_config_from_cmdline()
scripts/hostcfgd:1195: in init_kdump_config_from_cmdline
    kdump_config = self.config_db.get_entry("KDUMP", "config")
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <tests.common.mock_configdb.MockConfigDb object at 0x7f5117baec10>, key = 'KDUMP', field = 'config'

    def get_entry(self, key, field):
>       return MockConfigDb.CONFIG_DB[key][field]
E       KeyError: 'KDUMP'
```

vivekverma-arista:
Fixes: [#350](https://github.com/sonic-net/sonic-host-services/issues/350)

Refer the above issue for details.

Affected Tests

1. hostcfgd_fips_test.py
2. hostcfgd_logging_test.py
3. hostcfgd_rsyslog_test.py
4. hostcfgd_test.py

## How did I fix it? ##
1. The proposed fix is to properly mock cmdline for each of the tests instead of referring to the actual `/proc/cmdline` from the machine on which the test is running.

2. `crashkernel` has been removed from `/proc/cmdline` as all tests except `hostcfgd_test.py` expect kdump to be disabled.

3. Since only `hostcfd_test.py` expects kdump to be enabled, the parameter is explicitly injected to cmdline during setup phase of the test and reverted during teardown so that it doesn't affect other tests.


vivekverma-arista:
@qiluo-msft can you review?

PR #352, opened 2026-03-06
[gnoi-shutdown-daemon] Skip gNOI shutdown for already powered-off DPUs

### LINKED ISSUE #25889: Bug:[Smartswitch] Error log from gnoi-shutdown-daemon after setting admin state to down on a DPU which is powered off
### Is it platform specific

generic

### Importance or Severity

Critical

### Description of the bug

If DPU is currently powered off (with admin state set to down in configuration on config reload/ reboot commands will call the command `set_admin_state_gracefully` on an already powered off DPU. which will in-turn lead to gnoi-shutdown-daemon execution we get the error:
```
2026 Mar  4 16:57:47.694279 sonic ERR gnoi-shutdown-daemon[12171]: DPU0: Reboot command failed
2026 Mar  4 16:57:47.694376 sonic ERR gnoi-shutdown-daemon[12171]: DPU0: Failed to send Reboot command
```

### Steps to Reproduce

Have the configuration to power off the DPUs:
```
    "CHASSIS_MODULE": {
        "DPU0": {
            "admin_status": "down"
        },
        "DPU1": {
            "admin_status": "down"
        },
        "DPU2": {
            "admin_status": "down"
        },
        "DPU3": {
            "admin_status": "down"
        }
    },
```
Ensure DPUs are powered off
and then execute config reload / reboot:
The error logs are seen in the logs

```
2026 Mar  4 16:57:47.694279 sonic ERR gnoi-shutdown-daemon[12171]: DPU0: Reboot command failed
2026 Mar  4 16:57:47.694376 sonic ERR gnoi-shutdown-daemon[12171]: DPU0: Failed to send Reboot command
```

### Actual Behavior and Expected Behavior

No error logs should be seen on execution

### Relevant log output

```shell

```

```shell
202511_RC latest hash
```

### Attach files (if any)

_No response_

vvolam:
**What I did**

Added an operational status check in `gnoi_shutdown_daemon.py` `_handle_transition()` to skip the gNOI Reboot HALT sequence when the DPU is already not Online (e.g. Offline, PoweredDown).

**Why I did it**

Fixes https://github.com/sonic-net/sonic-buildimage/issues/25889

When DPUs are configured with `admin_status: "down"` and are already powered off, a config reload or reboot repopulates CONFIG_DB, which triggers `gnoi-shutdown-daemon` to attempt a gNOI Reboot HALT command on DPUs that are already offline — producing error logs:

```
ERR gnoi-shutdown-daemon[12171]: DPU0: Reboot command failed
ERR gnoi-shutdown-daemon[12171]: DPU0: Failed to send Reboot command
```

**How I verified it**

- Added unit tests for the new behavior:
  - `test_handle_transition_dpu_already_offline` — verifies skip when DPU is in `Offline` state
  - `test_handle_transition_dpu_powered_down` — verifies skip when DPU is in `PoweredDown` state
  - `test_handle_transition_oper_status_check_exception` — verifies graceful fallback when the oper status check raises an exception
- Updated existing tests to mock `get_oper_status()` returning `Online` so they continue testing the normal gNOI shutdown flow
- Manually verified the logic with mock tests

**Details if related**

The fix queries `get_oper_status()` via the platform chassis API at the start of `_handle_transition()`. If the DPU is not `Online`, the method logs a notice, clears the halt flag, and returns success — avoiding any gNOI RPC calls to unreachable DPUs. The check is wrapped in try/except so that if the platform API fails, the daemon falls back to the existing behavior.

### DIFF
--- a/scripts/gnoi_shutdown_daemon.py
+++ b/scripts/gnoi_shutdown_daemon.py
@@ -14,6 +14,7 @@ import os
 import redis
 import threading
 import sonic_py_common.daemon_base as daemon_base
+from sonic_platform_base.module_base import ModuleBase
 from sonic_py_common import syslogger
 from swsscommon import swsscommon
 
@@ -135,13 +136,18 @@ class GnoiRebootHandler:
                 module = self._chassis.get_module(module_index)
                 if module is not None:
                     oper_status = module.get_oper_status()
-                    if oper_status != "Online":
+                    if oper_status in (ModuleBase.MODULE_STATUS_OFFLINE,
+                                       ModuleBase.MODULE_STATUS_POWERED_DOWN):
                         logger.log_notice(
                             f"{dpu_name}: DPU is already in '{oper_status}' state, "
                             "skipping gNOI shutdown sequence"
                         )
-                        self._clear_halt_flag(dpu_name)
-                        return True
+                        cleared = self._clear_halt_flag(dpu_name)
+                        if not cleared:
+                            logger.log_warning(
+                                f"{dpu_name}: Failed to clear halt flag while skipping gNOI shutdown"
+                            )
+                        return cleared
         except Exception as e:
             logger.log_warning(
                 f"{dpu_name}: Could not determine operational status ({e}), "
--- a/scripts/gnoi_shutdown_daemon.py
+++ b/scripts/gnoi_shutdown_daemon.py
@@ -126,6 +126,28 @@ class GnoiRebootHandler:
         """
         logger.log_notice(f"{dpu_name}: Starting gNOI shutdown sequence")
 
+        # Check if DPU is already powered off / offline before attempting gNOI shutdown.
+        # This avoids error logs when config reload or reboot is issued while DPUs are
+        # already in the down state (e.g. admin_status was previously set to "down").
+        try:
+            module_index = self._chassis.get_module_index(dpu_name)
+            if module_index >= 0:
+                module = self._chassis.get_module(module_index)
+                if module is not None:
+                    oper_status = module.get_oper_status()
+                    if oper_status != "Online":
+                        logger.log_notice(
+                            f"{dpu_name}: DPU is already in '{oper_status}' state, "
+                            "skipping gNOI shutdown sequence"
+                        )
+                        self._clear_halt_flag(dpu_name)
+                        return True
+        except Exception as e:
+            logger.log_warning(
+                f"{dpu_name}: Could not determine operational status ({e}), "
+                "proceeding with gNOI shutdown"
+            )
+
         # Wait for platform PCI detach completion
         if not self._wait_for_gnoi_halt_in_progress(dpu_name):
             logger.log_warning(f"{dpu_name}: Timeout waiting for PCI detach, proceeding anyway")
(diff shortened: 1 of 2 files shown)


scripts/gnoi_shutdown_daemon.py
```diff
@@ -126,6 +126,28 @@ def _handle_transition(self, dpu_name: str, transition_type: str) -> bool:
...
+        # Check if DPU is already powered off / offline before attempting gNOI shutdown.
+        # This avoids error logs when config reload or reboot is issued while DPUs are
+        # already in the down state (e.g. admin_status was previously set to "down").
+        try:
+            module_index = self._chassis.get_module_index(dpu_name)
+            if module_index >= 0:
+                module = self._chassis.get_module(module_index)
+                if module is not None:
+                    oper_status = module.get_oper_status()
+                    if oper_status != "Online":
```
dgsudharsan: Can we use the macros MODULE_STATUS_ONLINE for checking here?

dgsudharsan: What if the module is at MODULE_STATUS_FAULT? Is it still ok to skip this call?

dgsudharsan (changes requested):

scripts/gnoi_shutdown_daemon.py
vvolam: I have addressed both the comments. Thank you.

dgsudharsan (approved):

scripts/gnoi_shutdown_daemon.py
```diff
@@ -126,6 +127,33 @@ def _handle_transition(self, dpu_name: str, transition_type: str) -> bool:
...
+                        logger.log_notice(
+                            f"{dpu_name}: DPU is already in '{oper_status}' state, "
+                            "skipping gNOI shutdown sequence"
+                        )
+                        cleared = self._clear_halt_flag(dpu_name)
+                        if not cleared:
+                            logger.log_warning(
+                                f"{dpu_name}: Failed to clear halt flag while skipping gNOI shutdown"
+                            )
+                        return cleared
```
hdwhdw: Consider extracting these logic into smaller functions like this, it avoid nesting and make it easier to read:
```python
def _should_skip_gnoi_shutdown(self, dpu_name: str) -> Optional[bool]:
    """
    Returns:
      True  -> DPU is known to be offline/powered down; skip gNOI shutdown.
      False -> DPU is known to be online (or some other state where we proceed).
      None  -> Cannot determine status; caller should proceed with gNOI shutdown.
    """
    module_index = self._chassis.get_module_index(dpu_name)
    if module_index < 0:
        return None

    module = self._chassis.get_module(module_index)
    if module is None:
        return None

    oper_status = module.get_oper_status()
    return oper_status in (
        ModuleBase.MODULE_STATUS_OFFLINE,
        ModuleBase.MODULE_STATUS_POWERED_DOWN,
    )
```

vvolam: Updated, please review

hdwhdw (approved):
LGTM

scripts/gnoi_shutdown_daemon.py
```diff
@@ -119,13 +120,60 @@ def __init__(self, db, config_db, chassis):
...
+                "proceeding with gNOI shutdown"
+            )
+            skip = False
+
+        if skip:
+            logger.log_notice(
+                f"{dpu_name}: DPU is already offline/powered-down, "
+                "skipping gNOI shutdown sequence"
+            )
+            cleared = self._clear_halt_flag(dpu_name)
```
gpunathilell, line 170: @vvolam  why do we need to clear the flag here?

vvolam: @gpunathilell We are returning from here if we are skipping the _handle_transition() and this is called with the flag set.

gpunathilell (approved):

PR #362, opened 2026-03-23
Fix test flakiness due to time.time() calls

### LINKED ISSUE #23745: Bug: Test flakiness in reboot_test.py due to delay in time.time() calls
### Is it platform specific

generic

### Importance or Severity

Medium

### Description of the bug

In the reboot_test.py, the assert calls check for the exact timestamp to match. However sometimes there is a delay between when the call is made and when the assert is checked. If this delay is greater than 1 second, the timestamps diverge and the assert fails. This is a flakiness introduced by the test framework and hence we should mock it and make it more robust.

### Steps to Reproduce

This is a hard to reproduce bug since it is heavily based on the timing delay. One can reproduce this by adding a sleep before the assert to simulate a delay of a few seconds.

### Actual Behavior and Expected Behavior

Actual behaviour: 
```
self = <MagicMock name='populate_reboot_status_flag' id='140170956134720'>
02:53:26  args = (False, 1774000374, 'Halt reboot did not complete', 3, <RebootStatus.STATUS_FAILURE: 3>)
02:53:26  kwargs = {}
02:53:26  expected = call(False, 1774000374, 'Halt reboot did not complete', 3, <RebootStatus.STATUS_FAILURE: 3>)
02:53:26  actual = call(False, 1774000373, 'Halt reboot did not complete', 3, <RebootStatus.STATUS_FAILURE: 3>)
```

Expected behaviour: Should not fail because of a timestamp mismatch.

### Relevant log output

```shell

```

```shell

```

### Attach files (if any)

_No response_

arista-hpandya:
Fixes: https://github.com/sonic-net/sonic-mgmt/issues/23745

In the reboot_test.py, the assert calls check for the exact timestamp to match. However sometimes there is a delay between when the call is made and when the assert is checked. If this delay is greater than 1 second, the timestamps diverge and the assert fails. This is a flakiness introduced by the test framework and hence we should mock it and make it more robust.

### DIFF
--- a/tests/host_modules/reboot_test.py
+++ b/tests/host_modules/reboot_test.py
@@ -204,13 +204,14 @@ class TestReboot(object):
         with (
             mock.patch("reboot._run_command") as mock_run_command,
             mock.patch("time.sleep") as mock_sleep,
+            mock.patch("time.time", return_value=TEST_TIMESTAMP),
             mock.patch("reboot.Reboot.populate_reboot_status_flag") as mock_populate_reboot_status_flag,
         ):
             mock_run_command.return_value = (0, ["stdout: execute WARM reboot"], ["stderror: execute WARM reboot"])
             self.reboot_module.execute_reboot("WARM")
             mock_run_command.assert_called_once_with("sudo warm-reboot")
             mock_sleep.assert_called_once_with(260)
-            mock_populate_reboot_status_flag.assert_called_once_with(False, int(time.time()), "Reboot command failed to execute", 'WARM', RebootStatus.STATUS_FAILURE)
+            mock_populate_reboot_status_flag.assert_called_once_with(False, int(TEST_TIMESTAMP), "Reboot command failed to execute", 'WARM', RebootStatus.STATUS_FAILURE)
 
     def test_execute_reboot_fail_unknown_reboot(self, caplog):
         with caplog.at_level(logging.ERROR):
@@ -221,6 +222,7 @@ class TestReboot(object):
     def test_execute_reboot_fail_issue_reboot_command_cold_boot(self, caplog):
         with (
             mock.patch("reboot._run_command") as mock_run_command,
+            mock.patch("time.time", return_value=TEST_TIMESTAMP),
             mock.patch("reboot.Reboot.populate_reboot_status_flag") as mock_populate_reboot_status_flag,
             caplog.at_level(logging.ERROR),
         ):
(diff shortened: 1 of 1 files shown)


rlhui:
@vvolam , please help review? thanks

tests/host_modules/reboot_test.py
```diff
@@ -265,6 +268,7 @@ def test_execute_reboot_fail_halt_timeout(self, caplog):
         with (
             mock.patch("reboot._run_command") as mock_run_command,
             mock.patch("time.sleep") as mock_sleep,
+            mock.patch("time.time", return_value=TEST_TIMESTAMP),
```
vvolam: Suggestion (pre-existing issue): `time.sleep` is mocked here but `time.monotonic()` is **not**. The production code uses `time.monotonic()` for the halt timeout loop (`while time.monotonic() - start_time < timeout`). Since sleep is a no-op but monotonic is real, this test spins for ~60 real wall-clock seconds waiting for the loop to exit.

Consider also mocking `time.monotonic` to make this test fast:

```suggestion
            mock.patch("time.time", return_value=TEST_TIMESTAMP),
            mock.patch("time.monotonic", side_effect=[0, 61]),
```

This makes the loop iterate once (0 -> start_time=0) then exit immediately on the second call (61 > 60 timeout). It would cut this test from ~60s to near-instant.

vvolam (commented):
Overall the `time.time()` fix looks good. One suggestion below for a related pre-existing issue in `test_execute_reboot_fail_halt_timeout`.

vvolam (commented):
@arista-hpandya changes LGTM, one small suggestion which you could fix along with the changes.

deepak-singhal0408:
@arista-hpandya could you check comments from @vvolam and also resolve conflicts.

rlhui:
> @arista-hpandya could you check comments from @vvolam and also resolve conflicts.

@arista-hpandya , @kenneth-arista  ping again on this pr

tests/host_modules/reboot_test.py
arista-hpandya: Thanks, implemented it.

arista-hpandya:
This PR (https://github.com/sonic-net/sonic-host-services/pull/391) merged recently that fixes the time mismatch issue, so the scope of this PR is now narrowed down to the timeout test optimization @vvolam suggested

judyjoseph (approved):

arista-hpandya:
@yxieca @rlhui @abdosi can we merge this?

vvolam (approved):

PR #367, opened 2026-03-30
Replace docker exec gnoi_client with native Python gRPC calls

sigabrtv1-ui:
## What I did

Refactored `gnoi_shutdown_daemon` to use direct Python gRPC calls instead of shelling out to `docker exec gnmi gnoi_client`.

## Why I did it

Three issues in the current subprocess-based approach:

1. **Broken RebootStatus detection** — the poll checks `"reboot complete" in out_s.lower()`, but `gnoi_client` outputs JSON (`{"active":false,"status":{"status":"STATUS_SUCCESS"}}`). The substring never matches, so **the poll always times out** regardless of DPU state.

2. **Suppressed error context** — `_send_reboot_command()` uses `suppress_stderr=True`, discarding Go panic traces from `gnoi_client`. The daemon logs only `"Reboot command failed"` with no actionable detail.

3. **Unnecessary NPU gnmi container dependency** — the current path shells into the NPU `gnmi` container to run `gnoi_client`, but the DPU's own gnmi server is the actual RPC endpoint. Direct gRPC connects without this intermediary.

## How I did it

- Vendored gNOI System proto Python stubs (from [openconfig/gnoi](https://github.com/openconfig/gnoi)) into `host_modules/gnoi/`
- Added a `GnoiClient` wrapper (`client.py`) with `reboot()`, `reboot_status()`, `cancel_reboot()` methods
- Refactored `_send_reboot_command` → `GnoiClient.reboot()` with proper `grpc.RpcError` handling (status codes + details)
- Refactored `_poll_reboot_status` → checks `resp.active` and `resp.status.status` protobuf fields directly
- Added `grpcio` and `protobuf` to `setup.py` dependencies
- Updated all unit tests + added new coverage for gRPC error paths

## How to verify it

On a SmartSwitch with DPU:
1. Set DPU admin_status to down in CONFIG_DB
2. Verify the daemon sends gNOI Reboot HALT via direct gRPC (no `docker exec` in logs)
3. Verify RebootStatus polling correctly detects completion (should NOT time out when DPU reports success)
4. Check syslog for structured error messages on failures (status code + details)

Design doc: #361


host_modules/gnoi/client.py
```diff
@@ -0,0 +1,99 @@
...
+Wraps the vendored proto stubs to provide reboot() and reboot_status()
+with proper error handling and structured responses.
+"""
+
+import grpc
+from host_modules.gnoi import system_pb2
+from host_modules.gnoi import system_pb2_grpc
+
+
+class GnoiClient:
```
hdwhdw, line 16: We might have other gnoi services. Should not pin it to System.

hdwhdw (changes requested):

host_modules/gnoi/client.py
sigabrtv1-ui: Good call — refactored in 5b12fba. `GnoiClient` now manages just the gRPC channel and exposes service stubs via properties (`client.system`). Future gNOI services (Healthz, Cert, etc.) can add their own `@property` or use `client.channel` directly for custom stubs.

hdwhdw: Call should be routed from through service. client.system.reboot instead of client.reboot.

sigabrtv1-ui: Updated in 9fa8000 — removed all convenience methods. Callers now go through `client.system.Reboot()` / `client.system.RebootStatus()` directly. GnoiClient just manages the channel + exposes service stubs via properties.

hdwhdw (approved):

PR #373, opened 2026-04-10
[featured] Fix namespace CONFIG_DBs rendering and cleanup

william8545:
## What I did

Fixed two issues in `featured` daemon for multi-ASIC platforms:

1. Feature deregistration now cleans up FEATURE entries from all per-ASIC namespace CONFIG_DBs and STATE_DBs.
2. `resync_feature_state()` now independently checks each namespace CONFIG_DB for unrendered Jinja2 templates, instead of assuming all namespaces match the host.

## How I did it

Deregistration cleanup: In `FeatureHandler.handler()`, when a feature is deregistered (`feature_cfg` is empty), added cleanup of the FEATURE entry from all namespace CONFIG_DBs via `set_entry(None)` and from all namespace STATE_DBs via `_del()`.

Independent namespace rendering: In `resync_feature_state()`, removed the early return when host state matches the rendered state. The host CONFIG_DB update still only happens for immutable or template states. Added a new loop that independently checks each namespace CONFIG_DB — if a namespace entry still contains a Jinja2 template, it gets rendered to the correct state value regardless of host CONFIG_DB state.

## How to verify it

On a multi-ASIC system:

```bash
# Test deregistration cleanup
sudo sonic-package-manager install xxx -y
sudo sonic-package-manager uninstall xxx -y --force
# Verify per-ASIC entries are removed
sonic-db-cli -n asic0 CONFIG_DB HGETALL "FEATURE|xxx"
# Should return {}

# Test template rendering
# After boot, check that per-ASIC FEATURE entries have rendered state values
sonic-db-cli -n asic0 CONFIG_DB HGET "FEATURE|xxx" state
```

### DIFF
--- a/scripts/featured
+++ b/scripts/featured
@@ -188,6 +188,13 @@ class FeatureHandler(object):
             syslog.syslog(syslog.LOG_INFO, "Deregistering feature {}".format(feature_name))
             self._cached_config.pop(feature_name, None)
             self._feature_state_table._del(feature_name)
+
+            # Clean up feature from namespace CONFIG_DBs and STATE_DBs
+            for ns, db in self.ns_cfg_db.items():
+                db.set_entry(FEATURE_TBL, feature_name, None)
+            for ns, tbl in self.ns_feature_state_tbl.items():
+                tbl._del(feature_name)
+
             return
 
         device_config = {}
@@ -526,9 +533,6 @@ class FeatureHandler(object):
         current_entry = self._config_db.get_entry('FEATURE', feature.name)
         current_feature_state = current_entry.get('state') if current_entry else None
 
-        if feature.state == current_feature_state:
-            return
-
         # feature.state might be rendered from a template, so that it should resync CONFIG DB
         # FEATURE table to override the template value to valid state value
         # ('always_enabled', 'always_disabled', 'disabled', 'enabled'). However, we should only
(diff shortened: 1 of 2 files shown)


oleksandrivantsiv (approved):

Sourabh-Kumar7:
@abdosi could you pls help with review?

tests/featured/featured_test.py
```diff
@@ -377,6 +377,86 @@ def test_enable_and_disable_feature_ExclusionList_skips_actions(self, mock_syslo
...
+        assert 'sflow' not in feature_handler._cached_config
+
+
+class TestResyncFeatureStateNamespace(TestCase):
+
+    @mock.patch('featured.FeatureHandler.update_systemd_config', mock.MagicMock())
+    @mock.patch('featured.FeatureHandler.update_feature_state', mock.MagicMock())
+    @mock.patch('featured.FeatureHandler.sync_feature_scope', mock.MagicMock())
+    @mock.patch('featured.FeatureHandler.sync_feature_delay_state', mock.MagicMock())
+    def test_namespace_template_rendered_when_host_matches(self):
```
qiluo-msft: None of the four patched methods are reachable, so the decorators are dead weight that makes the test noisier than it needs to be. Drop them.

scripts/featured
```diff
@@ -526,9 +533,6 @@ class FeatureHandler(object):
         current_entry = self._config_db.get_entry('FEATURE', feature.name)
         current_feature_state = current_entry.get('state') if current_entry else None
```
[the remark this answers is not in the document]
william8545: fix

tests/featured/featured_test.py
william8545: fix

Sourabh-Kumar7:
@abdosi @qiluo-msft could you please help re-review the change?

abdosi:
what is reason for making this change ?

qiluo-msft (approved):

william8545:
> what is reason for making this change ?

This is the featured-daemon half of a two-part fix for multi-ASIC platforms, where per-ASIC namespace FEATURE state diverged from the host for dynamically installed packages — config feature state <pkg> enabled failed, and uninstall left stale entries in the ASIC namespaces. The package-manager/CLI half is here: https://github.com/sonic-net/sonic-utilities/pull/4445 featured subscribes only to the host CONFIG_DB and is what propagates state into the namespaces, so three things can only be done here:
1. STATE_DB cleanup on deregister
2. Independent per-namespace rendering in resync_feature_state()
3. Resync on (re-)registration

PR #376, opened 2026-04-16
[smartswitch] Use dpu_halt_services_timeout from platform.json, fallback to HALT_TIMEOUT=60

rameshraghupathy:
### Description of PR
Summary:
On SmartSwitch platforms, HALT reboot may require additional time for DPU services and containers to terminate cleanly before the reboot sequence is considered complete.

The current implementation uses a fixed HALT_TIMEOUT=60 seconds, which may not be sufficient on all platforms and hardware combinations.

This change makes the HALT timeout configurable through platform.json using:

"dpu_halt_services_timeout"

while preserving the existing behavior by falling back to:

HALT_TIMEOUT = 60

when:

the key is missing,
the value is null,
the value is invalid,
the file cannot be read,
or the configured value is non-positive.

### Type of change
- [ x] Testbed and Framework(new/improvement)

### Back port request
- [ ] 202205
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411
- [ ] 202505
- [ ] 202511

#### How did you do it
Added helper function get_dpu_halt_services_timeout()
Read dpu_halt_services_timeout from platform.json
Added validation to ensure only positive timeout values are accepted
Added informational logging when fallback timeout is used
Updated HALT reboot flow to use platform-configured timeout
Added unit test coverage for:
missing key in platform.json
stable mocking of timeout helper in HALT reboot tests
Default behavior

If dpu_halt_services_timeout is not configured, the existing default behavior remains unchanged:

HALT_TIMEOUT = 60

#### How did you verify/test it?

normal HALT reboot flow
HALT timeout failure flow
missing platform.json key handling
invalid/unreadable timeout fallback handling
existing reboot UTs

### DIFF
--- a/host_modules/reboot.py
+++ b/host_modules/reboot.py
@@ -2,6 +2,7 @@
 
 import json
 import logging
+import os
 import threading
 import time
 import docker
@@ -9,6 +10,7 @@ import psutil
 from host_modules import host_service
 from utils.run_cmd import _run_command
 from enum import Enum
+from sonic_py_common import device_info
 
 MOD_NAME = 'reboot'
 # Reboot method in reboot request
@@ -33,6 +35,26 @@ class RebootStatus(Enum):
     STATUS_FAILURE = 3
 
 logger = logging.getLogger(__name__)
+
+
+def get_dpu_halt_services_timeout():
+    """Read dpu_halt_services_timeout from platform.json.
+       Fall back to HALT_TIMEOUT if missing/null/invalid/unreadable.
+    """
+    try:
+        platform_json_path = os.path.join(
+            device_info.get_path_to_platform_dir(), "platform.json"
+        )
+        with open(platform_json_path, "r") as f:
+            data = json.load(f)
+
+        timeout = data.get("dpu_halt_services_timeout")
+        if timeout is None:
+            return HALT_TIMEOUT
+
+        return int(timeout)
+    except (OSError, ValueError, TypeError, AttributeError, json.JSONDecodeError):
+        return HALT_TIMEOUT
 
 
 class Reboot(host_service.HostModule):
@@ -139,12 +161,12 @@ class Reboot(host_service.HostModule):
            is still alive after the below waiting period, we can conclude that the reboot has failed.
            Each container can take up to 20 seconds to get killed. In total, there are 10 containers,
            and adding a buffer of 1 minute brings up the delay value.
-           For Halt reboot_method, wait for 60 secs timeout. we expect pmon, syncd containers are killed, 
+           For Halt reboot_method, wait for configured timeout or fallback timeout. we expect pmon, syncd containers are killed,
            if Halt reboot is Successful."""
         if reboot_method in REBOOT_METHOD_HALT_BOOT_VALUES:
             # Periodically check every 5 seconds until PMON container is stopped or timeout occurs
             logger.info("%s: Waiting until services are halted or timeout occurs", MOD_NAME)
-            timeout = HALT_TIMEOUT
+            timeout = get_dpu_halt_services_timeout()
             start_time = time.monotonic()
 
             while time.monotonic() - start_time < timeout:
--- a/host_modules/reboot.py
+++ b/host_modules/reboot.py
@@ -52,8 +52,13 @@ def get_dpu_halt_services_timeout():
         if timeout is None:
             return HALT_TIMEOUT
 
-        return int(timeout)
-    except (OSError, ValueError, TypeError, AttributeError, json.JSONDecodeError):
+        timeout = int(timeout)
+        return timeout if timeout > 0 else HALT_TIMEOUT
+    except (OSError, ValueError, TypeError, AttributeError, json.JSONDecodeError) as e:
+        logger.info(
+            "%s: Failed to read dpu_halt_services_timeout from platform.json: %s. Using default %d",
+            MOD_NAME, e, HALT_TIMEOUT
+        )
         return HALT_TIMEOUT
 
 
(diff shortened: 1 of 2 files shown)


vvolam (changes requested):
@rameshraghupathy Changes are fine. Could you fix the coverage?

rameshraghupathy:
> @rameshraghupathy Changes are fine. Could you fix the coverage?

@vvolam Sure, will do.

vvolam:
@rameshraghupathy please fix PR description and use PR template

host_modules/reboot.py
```diff
@@ -35,6 +37,26 @@ class RebootStatus(Enum):
...
+            device_info.get_path_to_platform_dir(), "platform.json"
+        )
+        with open(platform_json_path, "r") as f:
+            data = json.load(f)
+
+        timeout = data.get("dpu_halt_services_timeout")
+        if timeout is None:
+            return HALT_TIMEOUT
+
+        return int(timeout)
```
vvolam: **2. No validation on timeout value**

`int(timeout)` can return 0 or negative values. A `platform.json` with `"dpu_halt_services_timeout": 0` or `-5` would cause the halt polling loop to exit immediately and falsely report success/failure.

Suggestion:
```python
val = int(timeout)
return val if val > 0 else HALT_TIMEOUT
```

host_modules/reboot.py
```diff
@@ -35,6 +37,26 @@ class RebootStatus(Enum):
...
+    except (OSError, ValueError, TypeError, AttributeError, json.JSONDecodeError):
+        return HALT_TIMEOUT
```
vvolam, line 62: **3. Silent fallback — no logging**

When `platform.json` is missing/unreadable/invalid, the function silently returns `HALT_TIMEOUT`. Adding a `logger.info()` in this `except` block would help operators diagnose timeout-related issues on the DPU.

Suggestion:
```python
except (OSError, ValueError, TypeError, AttributeError, json.JSONDecodeError) as e:
    logger.info("%s: Failed to read dpu_halt_services_timeout from platform.json: %s. Using default %d", MOD_NAME, e, HALT_TIMEOUT)
    return HALT_TIMEOUT
```

tests/host_modules/reboot_test.py
```diff
@@ -89,6 +90,44 @@ def test_populate_reboot_status_flag_with_status(self):
...
+
+    def test_get_dpu_halt_services_timeout_none(self):
+        mock_data = {"dpu_halt_services_timeout": None}
+
+        with (
+            mock.patch("host_modules.reboot.device_info.get_path_to_platform_dir", return_value="/tmp/platform"),
+            mock.patch("builtins.open", mock.mock_open(read_data=json.dumps(mock_data))),
+            mock.patch("host_modules.reboot.json.load", return_value=mock_data),
+        ):
+            assert host_reboot.get_dpu_halt_services_timeout() == HALT_TIMEOUT
```
vvolam, line 111: **4. Missing test: key absent from platform.json**

There's a test for `None` value but not for when the key is entirely missing from the JSON (e.g., `mock_data = {}`). While the code path is the same (`data.get()` → `None` → fallback), it's a distinct real-world scenario worth covering explicitly — it's the most common case on non-SmartSwitch platforms.

host_modules/reboot.py
```diff
@@ -139,12 +161,12 @@ def execute_reboot(self, reboot_method):
...
            Each container can take up to 20 seconds to get killed. In total, there are 10 containers,
            and adding a buffer of 1 minute brings up the delay value.
-           For Halt reboot_method, wait for 60 secs timeout. we expect pmon, syncd containers are killed, 
+           For Halt reboot_method, wait for configured timeout or fallback timeout. we expect pmon, syncd containers are killed,
            if Halt reboot is Successful."""
         if reboot_method in REBOOT_METHOD_HALT_BOOT_VALUES:
             # Periodically check every 5 seconds until PMON container is stopped or timeout occurs
             logger.info("%s: Waiting until services are halted or timeout occurs", MOD_NAME)
-            timeout = HALT_TIMEOUT
+            timeout = get_dpu_halt_services_timeout()
```
vvolam, line 174: **5. Existing halt tests are now fragile**

`test_execute_reboot_success_halt` and `test_execute_reboot_fail_halt_timeout` don't mock `get_dpu_halt_services_timeout()`. They currently work because there's no real `platform.json` in the test environment, so the function hits an exception and falls back to 60. If the test environment ever has a `platform.json` (e.g., running tests on an actual switch or in a build environment with platform packages installed), these tests would use an unexpected timeout value.

Suggestion: add `mock.patch("reboot.get_dpu_halt_services_timeout", return_value=60)` to both halt test methods.

vvolam (changes requested):
**1. PR description contradicts the code**

The description says "increasing the timeout to 120s for now" but the code reads `dpu_halt_services_timeout` from `platform.json` with a **60s fallback** — it doesn't hardcode 120. Please update the PR description to accurately explain the `platform.json` lookup mechanism and the fallback behavior. Also please use the PR template.

vvolam:
@rameshraghupathy could you address the comments?

host_modules/reboot.py
rameshraghupathy: @vvolam I will validate that the parsed timeout is positive; otherwise fallback to HALT_TIMEOUT.

host_modules/reboot.py
rameshraghupathy: @vvolam Added logger.info() when reading/parsing platform.json fails

tests/host_modules/reboot_test.py
rameshraghupathy: @vvolam Added an explicit test for missing key {}.

host_modules/reboot.py
rameshraghupathy: @vvolam Added mock get_dpu_halt_services_timeout() in the existing halt execute tests to avoid environment-dependent behavior.

vvolam (approved):
LGTM, Thanks

gpunathilell (approved):

PR #377, opened 2026-04-20
Fix device_metadata_handler to ignore non-localhost key data event

benle7:
What I did:
I updated device_metadata_handler in hostcfgd so that hostname, timezone, and rsyslog handling run only when the CONFIG_DB notification is for the localhost row. For any other DEVICE_METADATA key (for example bmc), the handler logs once at INFO and returns without calling hostname_update, timezone_update, or rsyslog_config.

Why I did it:
DEVICE_METADATA can have multiple keys; host settings live under localhost, while BMC fields live under bmc. The handler used to ignore the row key and always ran hostname logic, so bmc updates passed a dict without hostname, which produced ERR (“Hostname was not updated: Empty not allowed”).

How to verify
Change DEVICE_METADATA|bmc and confirm no "ERR Empty not allowed" in syslog.

### DIFF
--- a/tests/hostcfgd/hostcfgd_test.py
+++ b/tests/hostcfgd/hostcfgd_test.py
@@ -351,6 +351,27 @@ class TestHostcfgdDaemon(TestCase):
                 ]
                 mocked_syslog.syslog.assert_has_calls(expected)
 
+    def test_device_metadata_bmc_entry_does_not_trigger_hostname_update(self):
+        """
+        DEVICE_METADATA may include a 'bmc' row on BMC-capable systems. Those
+        updates must be ignored by device_metadata_handler so hostname_update
+        is not run with BMC-only fields (which lack hostname).
+        """
+        daemon = hostcfgd.HostConfigDaemon()
+        bmc_data = {
+            'bmc_addr': '169.254.0.1',
+            'bmc_if_addr': '169.254.0.2',
+            'bmc_if_name': 'usb0',
+            'bmc_net_mask': '255.255.255.252',
+        }
+        with mock.patch.object(daemon.devmetacfg, 'hostname_update') as mock_hostname_update:
+            with mock.patch.object(daemon.devmetacfg, 'timezone_update') as mock_tz_update:
+                with mock.patch.object(daemon.devmetacfg, 'rsyslog_config') as mock_rsyslog:
+                    daemon.device_metadata_handler('bmc', 'SET', bmc_data)
+                    mock_hostname_update.assert_not_called()
+                    mock_tz_update.assert_not_called()
+                    mock_rsyslog.assert_not_called()
+
     def test_mgmtiface_event(self):
         """
         Test handling mgmt events.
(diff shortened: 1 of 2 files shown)


Junchao-Mellanox (approved):

Sourabh-Kumar7:
@liuh-80 @Blueve can you pls help review this?

tests/hostcfgd/hostcfgd_test.py
```diff
@@ -351,6 +351,27 @@ def test_devicemeta_event(self):
...
+    def test_device_metadata_bmc_entry_does_not_trigger_hostname_update(self):
+        """
+        DEVICE_METADATA may include a 'bmc' row on BMC-capable systems. Those
+        updates must be ignored by device_metadata_handler so hostname_update
+        is not run with BMC-only fields (which lack hostname).
+        """
+        daemon = hostcfgd.HostConfigDaemon()
+        bmc_data = {
+            'bmc_addr': '169.254.0.1',
+            'bmc_if_addr': '169.254.0.2',
```
Blueve: Suggest make these mock value aligned with HLD to avoid misleading: https://github.com/sonic-net/SONiC/blob/master/doc/bmc/sonicBMC/pmon-bmc-design.md#213-host-bmc-link
```suggestion
            'bmc_addr': '169.254.100.1',
            'bmc_if_addr': '169.254.100.2',
```

benle7: Thanks @Blueve 

I found this issue with "169.254.0.1" because on the host side, each platform has its own bmc.json,
for example: sonic/device/x86_64-nvidia_sn6600_ld-r0/bmc.json.

In the HLD you added, it looks like there is a plan to create a global bmc.json file for the BMC side.

During the CR of https://github.com/sonic-net/sonic-buildimage/pull/26544 I asked 2 things:
- Leverage the global bmc.json file to be used on both sides: BMC and host.
- Use "169.254.0.1" IP address.

Do you think this PR (#377) can be approved or it is better to update the mock IP address to the new global address?

Blueve: I suggest make the test aligned with design. People might searching the codebase when they found a magic IP address. Use aligned IP could avoid potential misleading.

In the future, I think we might need a central place to record all special IP reservation inside SONiC to avoid conflicting happen between different features.

benle7: Resolved.
@Blueve Can you please approve?
Thanks!

benle7:
@Blueve Can you please approve?

Sourabh-Kumar7:
@Blueve could you please help re-review and approve this change?

Blueve (approved):

benle7:
@Sourabh-Kumar7 @Blueve Can you merge please?
Thanks!

rookie-who:
**Approved for 202605 cherry-pick.**

Must-have bug fix: device_metadata_handler processing non-localhost events causes incorrect behavior. PR has been tested on the 202605 branch and has code review approval on master.

*Review by @vaibhavhd via automated tooling*

PR #9, opened 2022-08-15
Support Restapi/gnmi control plane acls

ZhaohuiS:
Currently Sonic supports the following ACL tables.

IPV6_SNMP_ACL
EVERFLOW
SSH_ONLY
..

For the Restapi/gnmi use-cases, Sonic has to support a new Table: EXTERNAL_CLIENT of type CTRLPLANE, stage ingress

This shall match on 'src ip prefix' and dst port '8080'. Caclmgrd must parse this from acl.json and install as in the below example:

iptables -A INPUT -s 20.20.20.20/27 -p tcp --dport 8080 -j ACCEPT

or ip6tables if the 'src ip prefix' is IPv6.
**Steps:**

1. add EXTERNAL_CLIENT_ACL in config_db.json
```
"ACL_TABLE": {
        "NTP_ACL": {
            "policy_desc": "NTP_ACL",
            "services": [
                "NTP"
            ],
            "stage": "ingress",
            "type": "CTRLPLANE"
        },
        ...
       "EXTERNAL_CLIENT_ACL": {
            "policy_desc": "EXTERNAL_CLIENT_ACL",
            "services": [
                "EXTERNAL_CLIENT"
            ],
            "stage": "ingress",
            "type": "CTRLPLANE"
        }
}
```
2. config reload, then check show acl tables:
```
EXTERNAL_CLIENT_ACL  CTRLPLANE  EXTERNAL_CLIENT  EXTERNAL_CLIENT_ACL  ingress
NTP_ACL              CTRLPLANE  NTP              NTP_ACL              ingress
SNMP_ACL             CTRLPLANE  SNMP             SNMP_ACL             ingress
SSH_ONLY             CTRLPLANE  SSH              SSH_ONLY             ingress
```
3. load acl rules:
`acl-loader update full external_acl.json`
external_acl.json looks like this:

```
{
    "acl": {
        "acl-sets": {
            "acl-set": {
                "EXTERNAL_CLIENT-ACL": {
                    "acl-entries": {
                        "acl-entry": {
                            "1": {
                                "ip": {
                                    "config": {
                                        "source-ip-address": "20.0.0.3/32"
                                    }
                                }, 
                                "transport": {
				    "config": {
					"destination-port": "8081"
				    }
				},
                                "config": {
                                    "sequence-id": 1
                                }, 
                                "actions": {
                                    "config": {
                                        "forwarding-action": "ACCEPT"
                                    }
                                }
                            }
                        }
                    }, 
                    "config": {
[...]

PR #380, opened 2026-05-05
Added CACL entry for Redfish

shreyansh-nexthop:
#### Why I did it
The Redfish docker (docker-sonic-redfish) binds host port `0.0.0.0:443:18080` (per the docker-sonic-redfish rules in sonic-buildimage), so any IP that can reach the BMC management interface on TCP/443 can hit the Redfish API. `caclmgrd` only emits iptables rules for services that are registered in its `ACL_SERVICES` dict. Without an entry for REDFISH, any operator-defined ACL_TABLE referencing it is silently dropped i.e. the framework has no way to translate the operator's intent into iptables rules. This patch adds that registration and gates it so it only takes effect on platforms where Redfish actually runs.

#### How I did it
1. **Registration:** Added a REDFISH entry (tcp/443) to ACL_SERVICES in `scripts/caclmgrd`, mirroring the existing SSH/SNMP/NTP entries.
2. **Platform gating:** REDFISH only ships on BMC platforms, so to keep a stray services=["REDFISH"] template from programming tcp/443 rules elsewhere, REDFISH handling is gated on the FEATURE.redfish state:
    * A RedfishAllowed flag, checked in `get_acl_rules_and_translate_to_iptables_commands` - REDFISH services are skipped (with a warning) when the feature isn't enabled.
    * A FEATURE table subscriber flips the flag on runtime config feature state redfish enabled/disabled and re-walks so iptables converges immediately.
    * The flag is seeded from CONFIG_DB in __init__, before the initial iptables walk, so a caclmgrd restart doesn't leave tcp/443 momentarily ungated.
**Net:** REDFISH rules are programmed iff FEATURE.redfish.state == "enabled". On non-BMC platforms a stray REDFISH template is a no-op (warn + ignore).
3. **Tests:** Unit tests covering the walker for IPv4 + IPv6, in both enabled (rules emitted) and disabled (rules skipped) states.

#### How to verify it

1. Confirm the entry shipped:
   ```bash
   sudo grep -A4 '"REDFISH"' /usr/local/bin/caclmgrd
   ```

2. Confirm the feature is enabled (rules are only programmed when it is):
   ```bash
   sonic-db-cli CONFIG_DB hget "FEATURE|redfish" state    # expect: enabled
   ```

3. Create the CTRLPLANE table (acl-loader manages rules, not tables), then load rules via `acl-loader`:

   ```bash
   # Table (one-time; acl-loader does not create CTRLPLANE tables)
   redis-cli -n 4 HSET "ACL_TABLE|REDFISH_ACL" \
       policy_desc REDFISH_ACL type CTRLPLANE stage ingress services@ REDFISH

   # Rules
   cat > /tmp/redfish_acl.json <<'EOF'
   {
       "acl": {
           "acl-sets": {
               "acl-set": {
                   "REDFISH-ACL": {
                       "config": { "name": "REDFISH-ACL" },
                       "acl-entries": {
                           "acl-entry": {
                               "1": {
                                   "config": { "sequence-id": 1 },
                                   "ip": { "config": { "source-ip-address": "<your-source-ip>/32" } },
                                   "actions": { "config": { "forwarding-action": "ACCEPT" } }
                               }
                           }
                       }
                   }
               }
           }
       }
   }
   EOF

   acl-loader update full /tmp/redfish_acl.json
   ```

   > The acl-set name `REDFISH-ACL` maps to table `REDFISH_ACL` (uppercased, `-`→`_`). `acl-loader` auto-appends a `DEFAULT_RULE` (DROP), so the allowlist behaves as deny-by-default — no explicit deny rule needed. The destination port (443) comes from the `ACL_SERVICES` entry, so the rule only specifies the source IP.

4. Confirm caclmgrd translated them into iptables:
   ```bash
   show acl table REDFISH_ACL
   show acl rule REDFISH_ACL
   sudo iptables -L INPUT -n --line-numbers | grep ':443'
   ```
   You should see your `ACCEPT` for `<your-source-ip>` on dport 443 plus a trailing `DROP`.

Remove when done:
```bash
acl-loader delete REDFISH_ACL
redis-cli -n 4 DEL "ACL_TABLE|REDFISH_ACL"
```

### DIFF
--- a/scripts/caclmgrd
+++ b/scripts/caclmgrd
@@ -131,6 +131,7 @@ class ControlPlaneAclManager(logger.Logger):
     bfdAllowed = False
     VxlanAllowed = False
     VxlanSrcIP = ""
+    RedfishAllowed = False
     # a map from dpu name to port
     dashHaPortMap = {}
 
@@ -166,6 +167,9 @@ class ControlPlaneAclManager(logger.Logger):
         # Get all features that are present {feature_name : True/False}
         self.feature_present = {}
         self.update_feature_present()
+
+        feature_tb = self.config_db_map[DEFAULT_NAMESPACE].get_table(self.FEATURE_TABLE)
+        self.RedfishAllowed = feature_tb.get("redfish", {}).get("state") == "enabled"
 
         metadata = self.config_db_map[DEFAULT_NAMESPACE].get_table(self.DEVICE_METADATA_TABLE)
         if 'subtype' in metadata['localhost'] and metadata['localhost']['subtype'] == 'DualToR':
@@ -753,6 +757,11 @@ class ControlPlaneAclManager(logger.Logger):
                                      .format(table_name, acl_service))
                     continue
 
+                if acl_service == "REDFISH" and not self.RedfishAllowed:
+                    self.log_warning("Ignoring control plane ACL '{}': REDFISH service is not enabled on this platform"
+                                     .format(table_name))
+                    continue
+
                 self.log_info("Translating ACL rules for control plane ACL '{}' (service: '{}')"
                               .format(table_name, acl_service))
 
@@ -1059,6 +1068,14 @@ class ControlPlaneAclManager(logger.Logger):
         self.VxlanSrcIP = ""
         return True
 
+    def allow_redfish(self):
+        self.RedfishAllowed = True
+        self.log_info("Enabled REDFISH control plane ACL handling")
+
+    def block_redfish(self):
+        self.RedfishAllowed = False
+        self.log_info("Disabled REDFISH control plane ACL handling")
+
     def remove_dash_ha_rules(self, namespace, port):
         iptables_cmds = []
         # Remove iptables/ip6tables commands that allow DASH-HA packets
@@ -1164,6 +1181,9 @@ class ControlPlaneAclManager(logger.Logger):
 
         subscribe_dpu_table = swsscommon.SubscriberStateTable(config_db_connector, self.DPU_TABLE)
         sel.addSelectable(subscribe_dpu_table)
+
+        subscribe_feature_table = swsscommon.SubscriberStateTable(config_db_connector, self.FEATURE_TABLE)
+        sel.addSelectable(subscribe_feature_table)
         # Map of Namespace <--> susbcriber table's object
         config_db_subscriber_table_map = {}
 
@@ -1250,6 +1270,8 @@ class ControlPlaneAclManager(logger.Logger):
                         self.update_dhcp_acl(key, op, dict(fvs), mark)
                 continue
 
+            ctrl_plane_acl_notification = set()
+
             if db_id == config_db_id:
                 while True:
                     key, op, fvs = subscribe_vxlan_table.pop()
@@ -1267,7 +1289,21 @@ class ControlPlaneAclManager(logger.Logger):
                     if "dash-ha" in self.feature_present:
                         self.update_dash_ha_rules(namespace, key, op, fvs)
 
-            ctrl_plane_acl_notification = set()
+                while True:
+                    key, op, fvs = subscribe_feature_table.pop()
+                    if not key:
+                        break
+                    if key != "redfish":
+                        continue
+                    new_state = (op == 'SET' and dict(fvs).get("state") == "enabled")
+                    if new_state == self.RedfishAllowed:
+                        continue
+                    if new_state:
+                        self.allow_redfish()
+                    else:
+                        self.block_redfish()
+                    # Re-walk so any existing REDFISH ACL rules get (de)programmed now.
+                    ctrl_plane_acl_notification.add(namespace)
 
             # Pop data of both Subscriber Table object of namespace that got config db acl table event
             for table in config_db_subscriber_table_map[namespace]:
--- a/scripts/caclmgrd
+++ b/scripts/caclmgrd
@@ -1076,6 +1076,28 @@ class ControlPlaneAclManager(logger.Logger):
         self.RedfishAllowed = False
         self.log_info("Disabled REDFISH control plane ACL handling")
 
+    def handle_redfish_feature_events(self, subscribe_feature_table, namespace, ctrl_plane_acl_notification):
+        """Drain FEATURE table events and toggle REDFISH control plane ACL handling.
+
+        REDFISH iptables rules are only emitted when FEATURE.redfish is enabled. On a
+        state change, flip RedfishAllowed and request a re-walk so any existing REDFISH
+        ACL rules get (de)programmed immediately.
+        """
+        while True:
+            key, op, fvs = subscribe_feature_table.pop()
+            if not key:
+                break
+            if key != "redfish":
+                continue
+            new_state = (op == 'SET' and dict(fvs).get("state") == "enabled")
+            if new_state == self.RedfishAllowed:
+                continue
+            if new_state:
+                self.allow_redfish()
+            else:
+                self.block_redfish()
+            ctrl_plane_acl_notification.add(namespace)
+
     def remove_dash_ha_rules(self, namespace, port):
         iptables_cmds = []
         # Remove iptables/ip6tables commands that allow DASH-HA packets
@@ -1289,21 +1311,7 @@ class ControlPlaneAclManager(logger.Logger):
                     if "dash-ha" in self.feature_present:
                         self.update_dash_ha_rules(namespace, key, op, fvs)
 
-                while True:
-                    key, op, fvs = subscribe_feature_table.pop()
-                    if not key:
-                        break
-                    if key != "redfish":
-                        continue
-                    new_state = (op == 'SET' and dict(fvs).get("state") == "enabled")
-                    if new_state == self.RedfishAllowed:
-                        continue
-                    if new_state:
-                        self.allow_redfish()
-                    else:
-                        self.block_redfish()
-                    # Re-walk so any existing REDFISH ACL rules get (de)programmed now.
-                    ctrl_plane_acl_notification.add(namespace)
+                self.handle_redfish_feature_events(subscribe_feature_table, namespace, ctrl_plane_acl_notification)
 
             # Pop data of both Subscriber Table object of namespace that got config db acl table event
             for table in config_db_subscriber_table_map[namespace]:
--- a/scripts/caclmgrd
+++ b/scripts/caclmgrd
@@ -108,6 +108,11 @@ class ControlPlaneAclManager(logger.Logger):
             "dst_ports": ["22"],
             "multi_asic_ns_to_host_fwd":True
         },
+        "REDFISH": {
+            "ip_protocols": ["tcp"],
+            "dst_ports": ["443"],
+            "multi_asic_ns_to_host_fwd":False
+        },
         "EXTERNAL_CLIENT": {
             "ip_protocols": ["tcp"],
             "multi_asic_ns_to_host_fwd":False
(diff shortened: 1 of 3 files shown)


judyjoseph:
LGTM, @shreyansh-nexthop can you add sonic-mgmt test case to cover this.
Also can you refer may be this PR : https://github.com/sonic-net/sonic-host-services/pull/9 to check tests using acl-loader for this new ACL type ?

shreyansh-nexthop:
> LGTM, @shreyansh-nexthop can you add sonic-mgmt test case to cover this. Also can you refer may be this PR : #9 to check tests using acl-loader for this new ACL type ?

Sure @judyjoseph, will add sonic-mgmt test cases. Thanks!

tests/caclmgrd/test_redfish_acl_vectors.py
```diff
@@ -0,0 +1,77 @@
...
+                        "SRC_IP": "2001:db8::/32"
+                    },
+                },
+                "DEVICE_METADATA": {
+                    "localhost": {
+                    }
+                },
+                "FEATURE": {},
+            },
+            "return": [
```
ZhaohuiS, line 77: The test name says IPv6 src ip allow but uses `SRC_IP` (not `SRC_IPV6`). caclmgrd's `is_rule_ipv4()` checks the key name, so this will set `table_ip_version = 4` and emit `iptables` — not `ip6tables`. If the intent is to test real IPv6 enforcement, this should be `SRC_IPV6` with the expected output using `ip6tables`.

scripts/caclmgrd
```diff
@@ -108,6 +108,11 @@ class ControlPlaneAclManager(logger.Logger):
             "dst_ports": ["22"],
             "multi_asic_ns_to_host_fwd":True
         },
+        "REDFISH": {
```
ZhaohuiS, line 111: @shreyansh-nexthop I have a question, why introduce a new ACL table, not use EXTERNAL_CLIENT table instead?

tests/caclmgrd/test_redfish_acl_vectors.py
shreyansh-nexthop: Sure, make sense. The intent is to test IPv6, will do the required code changes. Thanks!

scripts/caclmgrd
shreyansh-nexthop: `EXTERNAL_CLIENT` has no dst_ports, while for `REDFISH` I want a fixed `dst_port` of 443. A dedicated REDFISH entry auto-scopes every rule to 443 (rather than every operator-written rule needing a manual --dport 443 match), and matches the existing SSH/SNMP/NTP pattern where each protocol gets its own ACL_SERVICES entry.

tests/caclmgrd/test_redfish_acl_vectors.py
shreyansh-nexthop: Done

ZhaohuiS (approved):

ZhaohuiS:
@shreyansh-nexthop could you pls run sonic-mgmt cacl related case to see if need to update them or add new cases to cover this new scenario?

shreyansh-nexthop:
> @shreyansh-nexthop could you pls run sonic-mgmt cacl related case to see if need to update them or add new cases to cover this new scenario?

@ZhaohuiS, the sonic-mgmt CACL testcases needs modification. I have done the code changes internally, currently testing it. Will soon raise the PR for it.

yxieca (changes requested):
Nice, tightly-scoped change overall — `caclmgrd` test pattern is followed cleanly and the PR description clears the description-as-merge-gate (root cause + fix + verification). One blocker plus a couple of questions before merge.

### Blocking

**1. `REDFISH` is platform-specific but `ACL_SERVICES` registration is unconditional (`scripts/caclmgrd:111-115`)**

All other entries in `ACL_SERVICES` (`SSH`, `NTP`, `SNMP`, `EXTERNAL_CLIENT`) describe services that exist on **every** SONiC platform — their `dst_ports` always correspond to a real listener. `REDFISH` is the first entry whose listener (bmcweb in `docker-sonic-redfish`) only ships on BMC-equipped platforms.

Today, on a non-BMC platform, an operator-defined `ACL_TABLE|REDFISH_ACL` with `services=["REDFISH"]` gets logged as `"unrecognized service"` and is silently dropped. After this PR, the same configuration on a non-BMC platform will:
- emit `iptables -A INPUT -p tcp --dport 443 ...` rules, and
- once `num_ctrl_plane_acl_rules > 0`, append the default-deny tail of the INPUT chain.

If anything else happens to listen on tcp/443 on that platform (future HTTPS REST, gRPC-TLS, NetConf-over-TLS, telemetry-over-HTTPS, a third-party agent), the operator's "Redfish ACL" template silently governs that traffic instead — or worse, locks them out. This is exactly the kind of fleet-template-meets-mixed-SKU foot-gun the framework should refuse to load itself into.

Please gate the REDFISH handling on platform/feature presence. Two reasonable shapes:

- **Feature-gated dispatch:** in `get_acl_rules_and_translate_to_iptables_commands` (or at `ACL_SERVICES` init), skip/warn for `REDFISH` unless `FEATURE.redfish` is enabled — analogous to how `bfdAllowed` / `VxlanAllowed` already gate other code paths in this same file.
- **Conditional registration:** only insert the REDFISH entry into `ACL_SERVICES` when the Redfish FEATURE / docker is present at daemon init.

Either makes `services=["REDFISH"]` on a non-BMC platform behave the same as it does today (warn + ignore), so a shared CTRLPLANE template across mixed SKUs doesn't accidentally program iptables on the wrong boxes.

### Please clarify

**2. Companion sonic-buildimage PR? (PR description)**

The body references "the docker-sonic-redfish rules in sonic-buildimage" that bind `0.0.0.0:443:18080`. Without that change, this entry is dormant; with it, the BMC-platform gating story above depends on how docker-sonic-redfish is FEATURE-gated. Please link the buildimage PR so we can verify the two changes ship together and the FEATURE table is set up the way the gating in (1) would check.

**3. Why a SONiC switch CACL for a BMC-hosted service?**

A one-line addition to the description ("on integrated chassis the BMC mgmt IP is reachable via the switch CPU mgmt path; this lets operators filter inbound 443 at the switch") would prevent reviewer confusion and make the platform-gating assumption explicit.

### Non-blocking nits

- `scripts/caclmgrd:111-115` — `multi_asic_ns_to_host_fwd: False` is the only False in the visible neighborhood. A one-line comment naming the assumption (Redfish container binds the host net on `0.0.0.0:443:18080`, no per-namespace forward needed) would save a future reader from chasing the buildimage docker config.
- `tests/caclmgrd/test_redfish_acl_vectors.py:65-76` — IPv6 vector covers ACCEPT only; the IPv4 vector covers ACCEPT + DROP. Adding a v6 DROP row exercises the `ip6tables ... -j DROP` translation path symmetrically.

Reviewed by AI agent on behalf of Ying.

shreyansh-nexthop:
> Nice, tightly-scoped change overall — `caclmgrd` test pattern is followed cleanly and the PR description clears the description-as-merge-gate (root cause + fix + verification). One blocker plus a couple of questions before merge.
> 
> ### Blocking
> **1. `REDFISH` is platform-specific but `ACL_SERVICES` registration is unconditional (`scripts/caclmgrd:111-115`)**
> 
> All other entries in `ACL_SERVICES` (`SSH`, `NTP`, `SNMP`, `EXTERNAL_CLIENT`) describe services that exist on **every** SONiC platform — their `dst_ports` always correspond to a real listener. `REDFISH` is the first entry whose listener (bmcweb in `docker-sonic-redfish`) only ships on BMC-equipped platforms.
> 
> Today, on a non-BMC platform, an operator-defined `ACL_TABLE|REDFISH_ACL` with `services=["REDFISH"]` gets logged as `"unrecognized service"` and is silently dropped. After this PR, the same configuration on a non-BMC platform will:
> 
> * emit `iptables -A INPUT -p tcp --dport 443 ...` rules, and
> * once `num_ctrl_plane_acl_rules > 0`, append the default-deny tail of the INPUT chain.
> 
> If anything else happens to listen on tcp/443 on that platform (future HTTPS REST, gRPC-TLS, NetConf-over-TLS, telemetry-over-HTTPS, a third-party agent), the operator's "Redfish ACL" template silently governs that traffic instead — or worse, locks them out. This is exactly the kind of fleet-template-meets-mixed-SKU foot-gun the framework should refuse to load itself into.
> 
> Please gate the REDFISH handling on platform/feature presence. Two reasonable shapes:
> 
> * **Feature-gated dispatch:** in `get_acl_rules_and_translate_to_iptables_commands` (or at `ACL_SERVICES` init), skip/warn for `REDFISH` unless `FEATURE.redfish` is enabled — analogous to how `bfdAllowed` / `VxlanAllowed` already gate other code paths in this same file.
> * **Conditional registration:** only insert the REDFISH entry into `ACL_SERVICES` when the Redfish FEATURE / docker is present at daemon init.
> 
> Either makes `services=["REDFISH"]` on a non-BMC platform behave the same as it does today (warn + ignore), so a shared CTRLPLANE template across mixed SKUs doesn't accidentally program iptables on the wrong boxes.
> 
> ### Please clarify
> **2. Companion sonic-buildimage PR? (PR description)**
> 
> The body references "the docker-sonic-redfish rules in sonic-buildimage" that bind `0.0.0.0:443:18080`. Without that change, this entry is dormant; with it, the BMC-platform gating story above depends on how docker-sonic-redfish is FEATURE-gated. Please link the buildimage PR so we can verify the two changes ship together and the FEATURE table is set up the way the gating in (1) would check.
> 
> **3. Why a SONiC switch CACL for a BMC-hosted service?**
> 
> A one-line addition to the description ("on integrated chassis the BMC mgmt IP is reachable via the switch CPU mgmt path; this lets operators filter inbound 443 at the switch") would prevent reviewer confusion and make the platform-gating assumption explicit.
> 
> ### Non-blocking nits
> * `scripts/caclmgrd:111-115` — `multi_asic_ns_to_host_fwd: False` is the only False in the visible neighborhood. A one-line comment naming the assumption (Redfish container binds the host net on `0.0.0.0:443:18080`, no per-namespace forward needed) would save a future reader from chasing the buildimage docker config.
> * `tests/caclmgrd/test_redfish_acl_vectors.py:65-76` — IPv6 vector covers ACCEPT only; the IPv4 vector covers ACCEPT + DROP. Adding a v6 DROP row exercises the `ip6tables ... -j DROP` translation path symmetrically.
> 
> Reviewed by AI agent on behalf of Ying.

Thanks @yxieca, addressed in the latest revision:

**(1) Blocking (platform gating):** REDFISH is now gated on `FEATURE.redfish` state — `RedfishAllowed` flag checked in the walker, flipped by a `FEATURE` table subscriber. On non-BMC platforms a `services=["REDFISH"]` template is a no-op (warn + ignore), same as today's "unrecognized service" path. Details in the updated PR description.

**(2) Companion buildimage PR:** https://github.com/sonic-net/sonic-buildimage/pull/27349 — adds Redfish as a feature and creates the `FEATURE.redfish` row.

**(3) Why a CACL for a BMC service:** the BMC runs its own SONiC instance, so `caclmgrd` and the Redfish docker are co-located on the BMC — caclmgrd filters inbound 443 on the BMC directly, same as it does for SSH/SNMP/NTP.

**Nits:** added the v6 DROP vector

yxieca (approved):
Round-2 review on 7fb36074: the cross-platform CACL concern from the prior review is properly addressed via FEATURE.redfish gating (init-time seed + walker skip + subscriber re-walk), with v4/v6 enabled+disabled test coverage and direct handler-branch tests. Nice work.

Reviewed by AI agent on behalf of Ying.

PR #384, opened 2026-05-12
[featured] fix stale FEATURE entry after concurrent package uninstall

DavidZagury:
**Why I did it**

After uninstalling a SONiC application package, the feature could reappear in show feature status with an incomplete entry (state: disabled, missing auto_restart, set_owner, etc.), even though the package was fully removed. This was caused by a race between featured and sonic-package-manager during uninstall.

sonic-package-manager unblocks as soon as featured writes disabled to STATE_DB, then immediately runs deregister() which deletes the FEATURE entry from CONFIG_DB. At that point featured is still inside its handler and has not finished writing back to CONFIG_DB. Two code paths independently recreate the deleted entry with partial data:

1. sync_feature_scope — called after a successful disable_feature(), it writes has_per_asic_scope and has_global_scope back to CONFIG_DB. disable_feature() writes disabled to STATE_DB as its last step, which immediately unblocks spm; by the time sync_feature_scope runs, spm may have already deleted the entry. mod_entry on a missing Redis key recreates the hash with only those two fields — no state key. featured then receives a notification for this partial entry, reads state=None, and logs ERR featured: Unexpected state value 'None' for feature <name>.
2. resync_feature_state — called when update_feature_state() fails (e.g. systemctl start fails because the service file was already removed by the concurrent uninstall). The enable path is triggered by config feature state <name> enabled followed immediately by sonic-package-manager uninstall. Spm unblocks on the earlier STATE_DB disabled written during install processing, removes the service file, and deletes the FEATURE entry — all while featured is still waiting for systemctl start to complete. When start fails, resync_feature_state is called and finds either a fully absent entry or the partial entry left by (1). In both cases current_feature_state is None, _feature_state_is_template(None) evaluates True, and mod_entry({state: disabled}) is called — completing or creating a ghost FEATURE record in CONFIG_DB that persists after the package is gone.

**How I did it**

Three guards added to featured:

- sync_feature_scope — post-write check: after both mod_entry calls, re-read the entry. If it exists but has no state key, the write raced with deregister and recreated a deleted key; delete the partial entry immediately.
- resync_feature_state — deleted-entry check: if get_entry() returns an empty dict, return without touching CONFIG_DB.
- resync_feature_state — partial-entry check: if the entry is non-empty but has no state key, it is the artifact from sync_feature_scope; delete it rather than completing the ghost record by writing the state field back.

Additionally, the ERR log for state=None in update_feature_state is suppressed since state=None means the FEATURE entry was deleted by a concurrent uninstall while the notification was queued — not a programming error. Any other unexpected non-None state still logs ERR.


vivekrnv (approved):

Sourabh-Kumar7:
@saiarcot895 could you please help review? Thanks.

Sourabh-Kumar7:
@DavidZagury can you please take a look at copilot's comments. Also please resolve the conflicts. thanks

DavidZagury:
> @DavidZagury can you please take a look at copilot's comments. Also please resolve the conflicts. thanks

@Sourabh-Kumar7 resolved

saiarcot895 (approved):

rookie-who:
Hi — this PR is flagged for cherry-pick to the **202605 branch** (`Request for 202605 Branch`). To move forward with the cherry-pick, could you please:

1. **Test this change on the 202605 branch** and add the test results to the PR description
2. **Add the label `Tested for 202605 branch`** once testing is complete

This helps us track readiness and ensures we don't cherry-pick untested changes. Thanks!

*Comment by @vaibhavhd via automated tooling*

rookie-who:
Hi — this PR has a cherry-pick conflict for the **202605 branch** (`Cherry Pick Conflict_202605`). Could you please rebase or resolve the conflict so we can proceed with the cherry-pick?

If you need help identifying the conflicting changes, let us know. Thanks!

*Comment by @vaibhavhd via automated tooling*

yxieca (approved):
## Review — Approve

The fix correctly addresses the stale `FEATURE` entry left behind when a package is uninstalled concurrently with feature resync. The guard "if the entry was deleted by a concurrent deregister, don't recreate it" plus the partial-entry cleanup is the right shape, and the tests exercise both the missing-entry and partial-entry (no `state` key) race paths.

### Findings (non-blocking)

- **Consistency (nit):** All new and existing `set_entry`/`get_entry`/`mod_entry` calls use the `FEATURE_TBL` constant — good; keep it that way (the 202605 backport PR #411 hardcodes `'FEATURE'` in one spot, which is functionally identical but breaks the constant convention). Nothing to change here in #384.
- The partial-entry deletion path relies on `set_entry(..., None)`; the added test asserting `set_entry` is called with `None` covers this well.

No blocking issues.

---
AI-assisted review, reviewed and approved by Ying Xie (@yxieca).

PR #387, opened 2026-06-01
Improve race condition handling in featured daemon-reload and docker container startup

arista-setu:
**Issue:**

This PR proposes an improvement for an open issue https://github.com/sonic-net/sonic-buildimage/issues/20988.

The underlying issue has been observed on chassis but is rarely seen and has not been reproducible. However, thorough code review suggests a potential race condition may exist between featured daemon-reload and docker container start process which is addressed by this PR

Based on current code review, following procedure during boot could have causes this issue:
- The featured daemon loops through every feature and for each one writes a systemd config file, calls systemctl daemon-reload (which temporarily tears down cgroup directories)
- Then, docker container starts 
- Here when daemon-reload fires for the next feature in loop ( here note that the container started in the previous iteration might still be initializing), it destroys the group path that the previous container's run process is trying to write its PID to, resulting in the "no such device" error.

https://github.com/sonic-net/sonic-host-services/blob/202405/scripts/featured#L227 > tiggers daemon-reload https://github.com/sonic-net/sonic-host-services/blob/202405/scripts/featured#L367

As a result, the docker start could fail as follow:
```
ERR container: docker cmd: start for eventd failed with 500 Server Error for http+docker://localhost/v1.43/containers/e6c2c0c3006061fd86d1736f8372ca4a10a7d8709d9a4c30c1da6d996576a2f0/start:
Internal Server Error ("failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: unable to apply cgroup configuration: failed to write 19294: write /sys/fs/cgroup/blkio/docker/e6c2c0c3006061fd86d1736f8372ca4a10a7d8709d9a4c30c1da6d996576a2f0/cgroup.procs: no such device: unknown")
```

**Proposed Improvement:**

The proposed fix will slightly change docker startup procedure. We could use 2 phase approach in `sync_state_field` : https://github.com/sonic-net/sonic-host-services/blob/202405/scripts/featured#L210) to avoid race condition between systemctl daemon-reload and Docker container startup:
- Phase 1 - write all systemd config files and trigger a single daemon-reload after all config is written
- Phase 2 - start all the containers 

**Testing:**

The issue was not reproduced before this change during testing.

However, I've attempted 5-6 reboots on chassis with 202405 image with this particular change. The underlying issue was still not reproduces and no other side-effects of this change were seen. All docker containers started as expected.

NOTE : Since that issue was never reliably reproduced, and other systemd units (started by sonic.target) actually may or may not cause the same race condition. The proposed improvement in featured is just a defensive approach to eliminate possibility that it race condition is cause by featured

**Backport request:**

I would like to request to be included in 202405. A direct backport will not apply to 202405 branch and hence a separate PR against the 202405 branch is create : https://github.com/sonic-net/sonic-host-services/pull/388

### DIFF
--- a/scripts/featured
+++ b/scripts/featured
@@ -221,7 +221,14 @@ class FeatureHandler(object):
         Summary:
         Updates the state field in the FEATURE|* tables as the state field
         might have to be rendered based on DEVICE_METADATA table and generated Device Running Metadata
+
+        Uses a two-phase approach to avoid a race condition between systemctl
+        daemon-reload and Docker container startup:
+        Phase 1: Write all systemd config files, followed by daemon-reload.
+        Phase 2: Start/stop all feature services (no daemon-reloads required).
         """
+        # Phase 1: Write all systemd configs and cache features
+        syslog.syslog(syslog.LOG_INFO, "Writing systemd configs")
         for feature_name in feature_table.keys():
             if not feature_name:
                 syslog.syslog(syslog.LOG_WARNING, "Feature is None")
@@ -233,7 +240,18 @@ class FeatureHandler(object):
             feature = Feature(feature_name, feature_table[feature_name], device_config)
 
             self._cached_config.setdefault(feature_name, feature)
-            self.update_systemd_config(feature)
+            self.update_systemd_config(feature, reload=False)
+
+        # Perform daemon-reload after all configs are written
+        self.reload_systemd_config()
+
+        # Phase 2: Start/stop services without daemon-reloads
+        syslog.syslog(syslog.LOG_INFO, "Restarting services")
+        for feature_name in feature_table.keys():
+            if not feature_name:
+                syslog.syslog(syslog.LOG_WARNING, "Feature is None")
+                continue
+            feature = self._cached_config[feature_name]
             self.update_feature_state(feature)
             self.sync_feature_scope(feature)
             self.resync_feature_state(feature)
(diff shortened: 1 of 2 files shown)


scripts/featured
```diff
@@ -233,7 +240,18 @@ class FeatureHandler(object):
             feature = Feature(feature_name, feature_table[feature_name], device_config)
 
             self._cached_config.setdefault(feature_name, feature)
-            self.update_systemd_config(feature)
+            self.update_systemd_config(feature, reload=False)
```
anamehra, line 243: while this change may improve handling from featured, it may not completely remove th erace condition as the services are started by sonic.target, correct? If thats the case, may we rephrase title/description as improvement not a fix?

tests/featured/featured_test.py
```diff
@@ -524,7 +524,6 @@ def test_feature_events(self, mock_syslog, get_runtime):
                         call(['sudo', 'systemctl', 'unmask', 'dhcp_relay.service'], capture_output=True, check=True, text=True),
                         call(['sudo', 'systemctl', 'enable', 'dhcp_relay.service'], capture_output=True, check=True, text=True),
                         call(['sudo', 'systemctl', 'start', 'dhcp_relay.service'], capture_output=True, check=True, text=True),
-                        call(['sudo', 'systemctl', 'daemon-reload'], capture_output=True, check=True, text=True),
```
anamehra, line 527: we remove daemon-reload from this path but could there be additional tests added to validate the changed flow?

mlok-nokia (approved):

scripts/featured
arista-setu: Thanks for raising this. Yes, your right. `sonic.target` orchestrates services beyond just featured, and other systemd units **_may_** still trigger daemon-reload or manipulate cgroups during container startup, reintroducing the same race -- which suggests that this PR is not a fix for mentioned issue.

However, since that issue was never reliably reproduced, and other systemd units (started by `sonic.target`) actually **_may or may not_** cause the same race condition if container instantiation is not too tight like it was in featured loop.

**Updated the title and description**
Given we're shipping a defensive improvement with featured, I've rephrased the title/description to reflect this more accurately.

Thanks

tests/featured/featured_test.py
arista-setu: I've added new unit test cases to test 2-phase approach proposed in the PR.

anamehra (approved):

abdosi (approved):

PR #389, opened 2026-06-02
[caclmgrd] Restrict FRR daemon loopback TCP ports to FRR user

nsoma26:
## Summary

Lock down zebra (TCP/2601) and the fpmsyncd<->zebra socket (TCP/2620) on the loopback interface so only the FRR user (UID 300) can connect; every other local UID is dropped.

- Per port the ACCEPT (uid-owner 300) rule is appended **before** the DROP so iptables matches FRR-internal traffic first.
- Rules are emitted in every namespace `caclmgrd` manages (host on single-ASIC, per-ASIC bgp container namespaces on multi-ASIC) since FRR runs in each of them with the same UID and ports.
- The helper is invoked from `get_acl_rules_and_translate_to_iptables_commands`, so the rules are reinstalled on every full programming pass: at boot, on `systemctl restart caclmgrd`, and after any CONFIG_DB `ACL_TABLE`/`ACL_RULE` change — no additional plumbing required for persistence.

### What changed

`scripts/caclmgrd`:
- New class constants `FRR_USER_UID = 300` and `FRR_LOOPBACK_TCP_DROP_PORTS = [2601, 2620]`.
- New helper `generate_frr_daemon_loopback_acl_commands(namespace)` that emits the ACCEPT-then-DROP pair per port, namespace-prefixed.
- Call site added in the main rule generator, right after the existing ip6tables NOTRACK rules and before the CONFIG_DB ACL lookup.

`tests/caclmgrd/`:
- `caclmgrd_frr_loopback_acl_test.py` — 5 tests covering rule presence, ACCEPT-before-DROP ordering, helper rule count, ASIC-namespace prefix, and a regression guard (`test_no_generic_output_drop_precedes_frr_accept`) that asserts no IPv4 OUTPUT-chain DROP/REJECT precedes the FRR ACCEPT rules.
- `test_frr_loopback_acl_vectors.py` — minimal CONFIG_DB test vector.

## Test plan

- [x] `python3 -m pytest tests/caclmgrd/` — `53 passed` (5 new + 48 pre-existing) in a Debian-bookworm container with `libswsscommon` + `sonic-py-common` installed.
- [x] `test_no_generic_output_drop_precedes_frr_accept` regression guard verified in-tree.
- [x] On-device verification: load on a single-ASIC and a multi-ASIC chassis, `iptables -S OUTPUT` shows the ACCEPT/DROP pairs for ports 2601 and 2620 in the host and in each per-ASIC bgp container namespace.
- [x] Functional: confirm `vtysh -d zebra` still works as the FRR user and is rejected for other UIDs; confirm fpmsyncd<->zebra connection stays up.

### DIFF
--- a/scripts/caclmgrd
+++ b/scripts/caclmgrd
@@ -125,6 +125,17 @@ class ControlPlaneAclManager(logger.Logger):
     }
     smartswitch_midplane_bridge_ip = "169.254.200.254"
 
+    # Restrict access to FRR daemon TCP ports on the loopback interface.
+    # FRR processes run as a dedicated user (UID 300); only that UID is
+    # permitted to connect to these ports over `lo`, everything else is
+    # dropped. iptables match order requires the ACCEPT (uid-owner) rule to
+    # be appended before the DROP for the same port.
+    FRR_USER_UID = 300
+    FRR_LOOPBACK_TCP_DROP_PORTS = [
+        2601,  # zebra VTY
+        2620,  # fpmsyncd <-> zebra
+    ]
+
     UPDATE_DELAY_SECS = 0.5
 
     DualToR = False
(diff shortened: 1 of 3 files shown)


nsoma26:
@ZhaohuiS @dgsudharsan Could you please take a look?

yyynini (approved):
lgtm

nsoma26:
@StormLiangMS Thank you for initiating the review.
@yyynini Thank you for the review and approval.

abdosi (approved):

PR #405, opened 2026-07-02
[process-reboot-cause] Save boot_id and device fields for DPU reboot cause

chartsai-nvidia:
#### Why I did it
DPU reboot-cause entries written to `CHASSIS_STATE_DB` were missing the `boot_id` and
`device` fields, so consumers couldn't correlate a DPU reboot with its boot instance or
source device. NPU entries are unaffected.

##### Work item tracking
- Microsoft ADO **(number only)**: N/A

#### How I did it
- Defined `NPU_REBOOT_CAUSE_DB_FIELDS` (cause, time, user, comment) and
  `DPU_REBOOT_CAUSE_DB_FIELDS` (adds boot_id, device) in `process-reboot-cause`.
- Replaced the hardcoded per-field `db.set` calls with a loop over the field set
  selected by device type, so DPU rows now include `boot_id` and `device`.
- Converted the showtech dump-path regex to a raw string to avoid invalid escape
  sequence warnings without changing its matching behavior.

#### How to verify it
- Run the unit tests: `tests/process-reboot-cause_test.py` now asserts that a DPU
  reboot-cause file results in `boot_id` and `device` being written to
  `CHASSIS_STATE_DB`.

- [ ] 201811
- [ ] 201911
- [ ] 202012
- [ ] 202106
- [ ] 202111
- [ ] 202205
- [ ] 202211
- [ ] 202305
- [ ] 202311
- [ ] 202405
- [ ] 202411

Save boot_id and device fields for DPU reboot-cause entries

### DIFF
--- a/scripts/process-reboot-cause
+++ b/scripts/process-reboot-cause
@@ -32,6 +32,8 @@ USER_ISSUED_REBOOT_CAUSE_REGEX ="User issued \'{}\' command [User: {}, Time: {}]
 REBOOT_CAUSE_UNKNOWN = "Unknown"
 REBOOT_CAUSE_TABLE_NAME = "REBOOT_CAUSE"
 MAX_HISTORY_FILES = 10
+DPU_REBOOT_CAUSE_DB_FIELDS = ("cause", "time", "user", "comment", "boot_id")
+NPU_REBOOT_CAUSE_DB_FIELDS = ("cause", "time", "user", "comment")
 
 REDIS_HOSTIP = "127.0.0.1"
 state_db = None
@@ -73,10 +75,10 @@ def read_reboot_cause_files_and_save_to_db(device='npu'):
                             sonic_logger.log_warning(f"Missing 'name' in reboot-cause file")
                             continue  # Skip this file
                         _hash = f"{REBOOT_CAUSE_TABLE_NAME}|{device.upper()}|{data['name']}"
-                    db.set(table, _hash, 'cause', data.get('cause', ''))
-                    db.set(table, _hash, 'time', data.get('time', ''))
-                    db.set(table, _hash, 'user', data.get('user', ''))
-                    db.set(table, _hash, 'comment', data.get('comment', ''))
+                    db_fields = (NPU_REBOOT_CAUSE_DB_FIELDS if device == 'npu'
+                                 else DPU_REBOOT_CAUSE_DB_FIELDS)
+                    for field in db_fields:
+                        db.set(table, _hash, field, data.get(field, ''))
                 except json.decoder.JSONDecodeError as je:
                     sonic_logger.log_error("Unable to process reload cause file {}: {}".format(x, je))
                     pass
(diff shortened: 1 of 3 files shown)


dgsudharsan (approved):

vvolam (commented):
Thanks for the cleanup — the loop-over-field-set refactor is clear and backward-compatible (NPU rows keep exactly `cause`/`time`/`user`/`comment`). One correctness gap inline regarding the `device` field, plus an optional nit below.

Nit (optional): to keep the two tuples from drifting, consider deriving one from the other:
```python
NPU_REBOOT_CAUSE_DB_FIELDS = ("cause", "time", "user", "comment")
DPU_REBOOT_CAUSE_DB_FIELDS = NPU_REBOOT_CAUSE_DB_FIELDS + ("boot_id", "device")
```

scripts/process-reboot-cause
```diff
@@ -32,6 +32,8 @@ USER_ISSUED_REBOOT_CAUSE_REGEX ="User issued \'{}\' command [User: {}, Time: {}]
 REBOOT_CAUSE_UNKNOWN = "Unknown"
 REBOOT_CAUSE_TABLE_NAME = "REBOOT_CAUSE"
 MAX_HISTORY_FILES = 10
+DPU_REBOOT_CAUSE_DB_FIELDS = ("cause", "time", "user", "comment", "boot_id")
```
vvolam: The PR title and description say this saves both `boot_id` **and** `device`, but `DPU_REBOOT_CAUSE_DB_FIELDS` only adds `boot_id` — `device` is never written to `CHASSIS_STATE_DB`. The unit test also only asserts `boot_id`, so the gap isn't caught.

Impact is low in practice because `show reboot-cause` (`fetch_data_from_db` in sonic-utilities) falls back to deriving the module from the key segment (`REBOOT_CAUSE|DPU0|...`) when the `device` field is absent. But as written, the PR doesn't do what its title/description claim.

Recommendation: add `"device"` to the tuple — it's trivial (`data.get('device', '')` already handles it, and the DPU cause file carries `device`) — and assert it in the test. If `device` is instead considered redundant with the key, please reconcile by updating the title/description/test.

chartsai-nvidia: Good point. Will add `"device"` to the tuple.

gpunathilell (approved):

dgsudharsan (approved):

gpunathilell (approved):

vvolam:
The updated `test_read_reboot_cause_files_and_save_to_db` nicely confirms DPU rows now include `boot_id`/`device`, but nothing locks in the *"NPU entries are unaffected"* contract from the PR description. Since the field selection now hinges on `device == 'npu'`, a future edit to `NPU_REBOOT_CAUSE_DB_FIELDS` could silently start writing DPU-only fields into `STATE_DB` without any test failing.

Consider adding a negative assertion in an NPU-focused test (where `read_reboot_cause_files_and_save_to_db('npu')` is invoked) so the boundary is enforced:

```python
# NPU path must NOT write DPU-only fields
written_fields = {call.args[2] for call in mock_db.set.call_args_list}
assert "boot_id" not in written_fields
assert "device" not in written_fields
```

Purely additive — nothing here blocks merge.

vvolam (approved):
LGTM, Thanks!

PR #407, opened 2026-07-09
[tacacs]: Add SSH TraceId authorization attribute support

Rod-Persky:
This pull request introduces support for a new `traceid_authorization` option for TACACS+ authorization, allowing the system to include an SSH TraceId in TACACS+ authorization requests. The changes ensure this option is configurable, has a sensible default, and is properly reflected in configuration templates and tests.

Related:
- PR [sonic-net/sonic-buildimage#27667](https://github.com/sonic-net/sonic-buildimage/pull/27667)
- HLD [sonic-net/SONiC#2358](https://github.com/sonic-net/SONiC/issues/2358)

New feature: TACACS+ TraceId Authorization

* Added support for the `traceid_authorization` option in TACACS+ configuration, defaulting to `on`, and updated the Jinja template (`tacplus_nss.conf.j2`) and sample output files to include this option and its documentation. [[1]](diffhunk://#diff-d426984e60c6cb68cd6e84bafdf5b6375e4656a3f8e115aa3db7eca769c93f12R37-R43) [[2]](diffhunk://#diff-e61f9541ce054b67003162f9b7ccfbd2e6b1cf80bd7e7929e92736ddc65bff99R27-R31) [[3]](diffhunk://#diff-7ec9cc919e3a290c5d1b9ad5fed06edcbc9475eedb733865b52d1891eba34d0fR25-R29)
* Introduced the `TACPLUS_TRACEID_AUTHORIZATION_DEFAULT` constant (default: `True`) and included it in the global TACACS+ defaults in `hostcfgd`. [[1]](diffhunk://#diff-f46602bac3380a0d28ddfcbe698d95d87c7bc57e504fac08dd6906bdb31ec388R90) [[2]](diffhunk://#diff-f46602bac3380a0d28ddfcbe698d95d87c7bc57e504fac08dd6906bdb31ec388L369-R371)
* Ensured that the `traceid_authorization` value is correctly parsed and normalized when updating the global TACACS+ configuration.
* Updated the configuration rendering logic to pass the `traceid_authorization` value to the template.

Test updates

* Modified test vectors and sample outputs to cover the new `traceid_authorization` option, ensuring correct behavior when the option is set or unset. [[1]](diffhunk://#diff-ce9a748e51a4dde78ed6e8ffed42dfad301b09bb50657c6e898eacb24226a417L50-R51) [[2]](diffhunk://#diff-ce9a748e51a4dde78ed6e8ffed42dfad301b09bb50657c6e898eacb24226a417L112-R114)


scripts/hostcfgd
```diff
@@ -87,6 +87,7 @@ LIMITS_CONF = "/etc/security/limits.conf"
 TACPLUS_SERVER_PASSKEY_DEFAULT = ""
 TACPLUS_SERVER_TIMEOUT_DEFAULT = "5"
 TACPLUS_SERVER_AUTH_TYPE_DEFAULT = "pap"
+TACPLUS_TRACEID_AUTHORIZATION_DEFAULT = True
```
qiluo-msft: same opt-out concern as the `default true` in sonic-buildimage#27667 (these two are paired). when `tacacs_global_update` merges `tacplus_global_default` into `tacplus_global`, any existing CONFIG_DB entry without `traceid_authorization` gets `True` filled in — so `traceid_authorization` is written to `tacplus_nss.conf` and `TRACE_ID_AUTHORIZATION_FLAG` is set on every upgrade, silently. should be `False` to match opt-in semantics. the template `# Default: on` comment (tacplus_nss.conf.j2 line 40) would need updating too.

tests/hostcfgd/test_tacacs_vectors.py
```diff
@@ -47,7 +47,8 @@
                             "auth_type": "chap",
                             "timeout": 5,
                             "passkey": "dellsonic",
-                            "src_intf": "Ethernet0"
+                            "src_intf": "Ethernet0",
+                            "traceid_authorization": "false"
```
qiluo-msft, line 51: both test vectors set `"traceid_authorization": "false"` explicitly. there's no test for the "key absent from CONFIG_DB" path — i.e., what happens when an existing config has no `traceid_authorization` at all. that's the most important case for the backward-compat concern: it would use `TACPLUS_TRACEID_AUTHORIZATION_DEFAULT` and produce `traceid_authorization` in the rendered conf. if the default changes to `False`, the `TACACS_config_db_local_and_tacacs` sample output would also need updating (it now shows `traceid_authorization` active).

tests/hostcfgd/test_tacacs_vectors.py
```diff
@@ -109,7 +110,8 @@
                             "auth_type": "chap",
                             "timeout": 5,
                             "passkey": "dellsonic",
-                            "src_intf": "Ethernet0"
+                                "src_intf": "Ethernet0",
+                                "traceid_authorization": "false"
```
qiluo-msft: nit: the new `src_intf` and `traceid_authorization` lines have 4 extra spaces compared to the surrounding dict keys. python is fine with it but inconsistent.

scripts/hostcfgd
Rod-Persky: @qiluo-msft I've now changed this to opt-in rather than opt-out across sonic-host-services

tests/hostcfgd/test_tacacs_vectors.py
Rod-Persky: Fixed, sorry about that

tests/hostcfgd/test_tacacs_vectors.py
Rod-Persky: So we've now got three test cases,

Explicit false: [config_db_local](https://github.com/Rod-Persky/sonic-host-services/tree/dev/rodpersky/traceid-hld-support/tests/hostcfgd/sample_output/TACACS_config_db_local) set [here](https://github.com/Rod-Persky/sonic-host-services/blob/dev/rodpersky/traceid-hld-support/tests/hostcfgd/test_tacacs_vectors.py#L51C29-L51C61)
Explicit true: [config_db_tacacs](https://github.com/Rod-Persky/sonic-host-services/tree/dev/rodpersky/traceid-hld-support/tests/hostcfgd/sample_output/TACACS_config_db_tacacs) set [here](https://github.com/Rod-Persky/sonic-host-services/blob/dev/rodpersky/traceid-hld-support/tests/hostcfgd/test_tacacs_vectors.py#L114)
Key absent: [config_db_local_and_tacacs](https://github.com/Rod-Persky/sonic-host-services/tree/dev/rodpersky/traceid-hld-support/tests/hostcfgd/sample_output/TACACS_config_db_local_and_tacacs) absent [here](https://github.com/Rod-Persky/sonic-host-services/blob/dev/rodpersky/traceid-hld-support/tests/hostcfgd/test_tacacs_vectors.py#L172)


And then there is some additional files that changed simply because the j2 changes that repeat the key absent tests.

PR #409, opened 2026-07-14
[console-monitor] Implement Console Logging feature

nats-nokia:
console-monitor DCE reads console logging filename, logrotate parameters from CONSOLE_PORT table and creates logrotate conf for corresponding line number. If filename and logrotate are not provided, default values will be used.
 
console-monitor proxy service opens the log file name in append mode and keep logging the characters read from console device, if logging is enabled.


scripts/console-monitor
```diff
@@ -1068,17 +1147,73 @@ class DCEService:
...
+            for link_id in self._logrotate_links - desired_links:
+                conf_path = logrotate_conf_path(link_id)
+                try:
+                    if os.path.exists(conf_path):
+                        os.remove(conf_path)
+                        log.info(f"DCE: [{link_id}] Removed logrotate config: {conf_path}")
+                except OSError as e:
+                    log.error(f"DCE: [{link_id}] Failed to remove logrotate config {conf_path}: {e}")
+
+            self._logrotate_links = desired_links
```
ediwibowo-msft: The current logic only removes configs tracked in the in-memory self._logrotate_links set, which is empty after a restart. The fix is to reconcile against what's actually on disk. After a DCE restart, existing_links is rebuilt from the actual /etc/logrotate.d/console-* files, so any file for a port that has since been disabled or deleted from `CONSOLE_PORT` gets removed, regardless of whether this process ever tracked it in memory. One caveat to note: the glob-based approach reconciles the entire console-* namespace, so it assumes console-monitor owns every `LOGROTATE_CONF_PREFIX` file in `LOGROTATE_DIR`. 

I would suggest the change to `_sync_logrotate_configs`:
```
import glob
...
          # Reconcile against configs actually present on disk, not just the
          # in-memory set, so stale files left by a previous DCE run are cleaned up.
          existing_links = set()
          pattern = os.path.join(LOGROTATE_DIR, f"{LOGROTATE_CONF_PREFIX}*")
          for conf_path in glob.glob(pattern):
              existing_links.add(os.path.basename(conf_path)[len(LOGROTATE_CONF_PREFIX):])
          
          for link_id in existing_links - desired_links:
              conf_path = logrotate_conf_path(link_id)
              try:
                  os.remove(conf_path)
                  log.info(f"DCE: [{link_id}] Removed logrotate config: {conf_path}")
              except OSError as e:
                  log.error(f"DCE: [{link_id}] Failed to remove logrotate config {conf_path}: {e}")
          
          self._logrotate_links = desired_links
```

scripts/console-monitor
```diff
@@ -696,6 +738,18 @@ class ProxyService:
             # Open PTM
             self.ptm_fd = os.open(self.ptm_path, os.O_RDWR | os.O_NOCTTY | os.O_NONBLOCK)
 
+            if self.logging_enabled and self.log_file_path:
+                try:
+                    self.log_fd = os.open(
+                        self.log_file_path,
+                        os.O_WRONLY | os.O_CREAT | os.O_APPEND,
+                        0o644,
```
ediwibowo-msft: Console logs capture everything typed/echoed at the console, including credentials entered at login/enable prompts. `0o644` lets any local user read them. Consider `0o640` or `0o600`.

nats-nokia: sure, corrected the file permission to 0o640. Thank you.

scripts/console-monitor
nats-nokia: Though the actual log file would not be written in case of logging is disabled. It is good to remove the stale logrotate conf file, if present.
Addressed this issue. Thank you

scripts/console-monitor
```diff
@@ -490,6 +500,31 @@ def calculate_filter_timeout(baud: int, multiplier: int = 3) -> float:
...
+    """Return the default console log file path for a line."""
+    return DEFAULT_LOG_FILE_TEMPLATE.format(link_id)
+
+
+def build_logrotate_config(log_file: str, size: str, count: int) -> str:
+    """Build logrotate configuration for a console log file."""
+    return (
+        f"{log_file} {{\n"
+        f"    missingok\n"
+        f"    notifempty\n"
```
judyjoseph: should we have some validation checks for these arguments before making the log rotate config? or is it assumed that yang validation will catch it ?

nats-nokia: Yes, yang validation would catch any invalid arguments.

scripts/console-monitor
```diff
@@ -1068,17 +1148,81 @@ class DCEService:
...
         for link_id in redis_ids & current_ids:
             new_config = redis_configs[link_id]
             old_config = self._config_cache.get(link_id, {})
             if new_config != old_config:
                 log.info(f"DCE: [{link_id}] Config changed: {old_config} -> {new_config}")
+                if (old_config.get("logging_enabled") != new_config.get("logging_enabled") or
+                        old_config.get("log_file") != new_config.get("log_file") or
+                        old_config.get("logrotate_size") != new_config.get("logrotate_size") or
+                        old_config.get("logrotate_count") != new_config.get("logrotate_count")):
+                    log.warning(f"DCE: [{link_id}] Logging config changed, proxy will restart")
```
ediwibowo-msft: Changing only `logrotate_size`/`logrotate_count` will bounce the proxy and drop the active console session, even though the proxy's `_load_config` never reads these values (only `baud`, `logging_enabled`, `log_file`). Because `_get_all_configs` now includes the logrotate params in the cached dict, the outer `new_config != old_config` check restarts the link for a logrotate-only edit — but `_sync_logrotate_configs` already regenerates the conf, which is all that's needed. Consider keeping logrotate-only params out of the restart-detection config so those edits only rewrite the logrotate file without restarting the proxy. (No test currently asserts a logrotate-only change avoids restart.)

scripts/console-monitor
```diff
@@ -1068,17 +1148,81 @@ class DCEService:
...
+                    log.info(f"DCE: [{link_id}] Updated logrotate config: {conf_path}")
+                except OSError as e:
+                    log.error(f"DCE: [{link_id}] Failed to write logrotate config {conf_path}: {e}")
+
+            # Reconcile against configs actually present on disk, not just the
+            # in-memory set, so stale files left by a previous DCE run are cleaned up.
+            # This glob-based reconciliation assumes console-monitor owns every
+            # LOGROTATE_CONF_PREFIX file under LOGROTATE_DIR.
+            existing_links = set()
+            pattern = os.path.join(LOGROTATE_DIR, f"{LOGROTATE_CONF_PREFIX}*")
```
ediwibowo-msft: This reconciliation deletes any `/etc/logrotate.d/console-*` file not in the desired set. The `console-` prefix is generic, and as the comment above acknowledges, this assumes console-monitor owns every matching file. If another package ever ships `/etc/logrotate.d/console-<something>`, it would be silently removed. A more specific prefix (e.g. `console-monitor-` or `sonic-console-`) would make ownership unambiguous and keep this safe.

scripts/console-monitor
```diff
@@ -1034,13 +1108,19 @@ class DCEService:
...
+                    "logrotate_size": entry.get("logrotate_size", DEFAULT_LOGROTATE_SIZE),
+                    "logrotate_count": str(entry.get("logrotate_count", DEFAULT_LOGROTATE_COUNT)),
                 }
         except Exception as e:
             log.error(f"DCE: Failed to get configs: {e}")
         return configs
 
     def _sync(self) -> None:
         """Sync services with CONFIG_DB"""
+        self._sync_logrotate_configs()
```
ediwibowo-msft: `_sync_logrotate_configs()` runs before `_check_feature_enabled()`, so logrotate confs are written even when `console_mgmt` is disabled and no proxy is running. Harmless due to `missingok`, but inconsistent — when the feature is off there's nothing to rotate, and arguably the confs should be cleaned up rather than created.

scripts/console-monitor
```diff
@@ -696,6 +739,18 @@ class ProxyService:
             # Open PTM
             self.ptm_fd = os.open(self.ptm_path, os.O_RDWR | os.O_NOCTTY | os.O_NONBLOCK)
 
+            if self.logging_enabled and self.log_file_path:
+                try:
+                    self.log_fd = os.open(
```
ediwibowo-msft: `log_file_path` comes straight from CONFIG_DB and is opened as root with `O_CREAT|O_APPEND`, following symlinks to an arbitrary path. CONFIG_DB is admin-trusted so risk is low, but since this runs with elevated privileges it'd be worth hardening — e.g. confining the path to `/var/log` or adding `O_NOFOLLOW` — to avoid a stray symlink redirecting root writes.

ediwibowo-msft (commented):
Reviewed the console logging feature. Overall design is solid — `O_APPEND` + `copytruncate` is the right combination so the proxy fd survives rotation without reopening, and logging device output rather than raw PTM input sensibly avoids capturing un-echoed input. A few points below, one functional.

scripts/console-monitor
nats-nokia: Addressed. changing only logrotate parameters will not restart proxy.

scripts/console-monitor
nats-nokia: conf prefix changed to /var/log/console-monitor-logging-*.

scripts/console-monitor
nats-nokia: _sync_logrotate_configs will remove all conf if console_mgmt feature is disabled.

scripts/console-monitor
nats-nokia: Added O_NOFOLLOW flag.
Pls let me know if you want to force the log dir as only "/var/log/"

ediwibowo-msft: Adding O_NOFOLLOW is good enough.

PR #411, opened 2026-07-16
[202605][featured] fix stale FEATURE entry after concurrent package uninstall

DavidZagury:
Cherry-pick of #384 onto the 202605 branch.

**Why I did it**

After uninstalling a SONiC application package, the feature could reappear in `show feature status` with an incomplete entry (state: disabled, missing auto_restart, set_owner, etc.), even though the package was fully removed. This was caused by a race between `featured` and `sonic-package-manager` during uninstall.

`sonic-package-manager` unblocks as soon as `featured` writes `disabled` to STATE_DB, then immediately runs `deregister()` which deletes the FEATURE entry from CONFIG_DB. At that point `featured` is still inside its handler and has not finished writing back to CONFIG_DB. Two code paths independently recreate the deleted entry with partial data:

1. `sync_feature_scope` — called after a successful `disable_feature()`, it writes `has_per_asic_scope` and `has_global_scope` back to CONFIG_DB. `mod_entry` on a missing Redis key recreates the hash with only those two fields (no `state` key). `featured` then receives a notification for this partial entry, reads `state=None`, and logs `ERR featured: Unexpected state value 'None' for feature <name>`. There is also a TOCTOU gap: spm can delete the entry between the `mod_entry` calls and the post-write read.

2. `resync_feature_state` — called when `update_feature_state()` fails (e.g. `systemctl start` fails because the service file was already removed by the concurrent uninstall). `get_entry()` returns `{}` for a deleted key, `_feature_state_is_template(None)` evaluates True, and `mod_entry` writes `{state: disabled}` back — recreating the entry with no other fields. The same function also completes partial entries left by (1).

**How I did it**

Three guards added to `featured`:

- `sync_feature_scope` — post-write check: after both `mod_entry` calls, re-read the entry. If it exists but has no `state` key, the write raced with deregister and recreated a deleted key; delete the partial entry immediately.
- `resync_feature_state` — deleted-entry check: if `get_entry()` returns an empty dict, return without touching CONFIG_DB.
- `resync_feature_state` — partial-entry check: if the entry is non-empty but has no `state` key, delete it rather than completing the ghost record.

Additionally, the ERR log for `state=None` in `update_feature_state` is suppressed since `state=None` means the FEATURE entry was deleted by a concurrent uninstall while the notification was queued — not a programming error.

**Cherry-pick notes**

Picked cleanly from #384 (commits 8c1737f, b10757c, ad26790, 800288d) with no conflicts.

### DIFF
--- a/scripts/featured
+++ b/scripts/featured
@@ -266,8 +266,12 @@ class FeatureHandler(object):
             return False
 
         if not enable and not disable:
-            syslog.syslog(syslog.LOG_ERR, "Unexpected state value '{}' for feature {}"
-                          .format(feature.state, feature.name))
+            # state=None means the FEATURE entry was deleted by a concurrent uninstall
+            # while this notification was queued — not an error.  Any other non-None
+            # value here is a genuinely unexpected state and should be flagged.
+            if feature.state is not None:
+                syslog.syslog(syslog.LOG_ERR, "Unexpected state value '{}' for feature {}"
+                              .format(feature.state, feature.name))
             return False
 
         if feature.delayed and not self.is_delayed_enabled:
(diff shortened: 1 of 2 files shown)


yxieca (approved):
## Review — Approve (202605 backport)

Faithful backport of #384 (merged on master). The core fix — guarding against recreating a `FEATURE` entry deleted by a concurrent deregister, plus partial-entry cleanup — and the accompanying tests match the original.

### Findings (non-blocking)

- **Consistency (nit):** One call uses the hardcoded literal `set_entry('FEATURE', feature.name, None)` where the master version (#384) uses the `FEATURE_TBL` constant. These are functionally identical (`FEATURE_TBL == swsscommon.CFG_FEATURE_TABLE_NAME == "FEATURE"`), and every other call in this file uses the constant, so aligning it to `FEATURE_TBL` would keep the convention consistent. Not blocking for a backport.

No functional or behavioral differences from the merged master fix. No blocking issues.

---
AI-assisted review, reviewed and approved by Ying Xie (@yxieca).

PR #414, opened 2026-07-26
Add console mirror recording and control support

V2lsbGlhbVpoYW5n:
#### What I did

Implemented the host-service component of the Console Mirror feature described
in [HLD](https://github.com/sonic-net/SONiC/pull/2388).

#### Why I did it

Operators need an on-device way to record console traffic for troubleshooting
without interrupting the existing serial forwarding path.

#### How I did it

- Added `host_modules/console_mirror.py`:
- Updated `console-monitor proxy` to:
  - create a mirror manager and control server for each proxy instance
  - submit RX and TX payloads after successful forwarding

### DIFF
--- a/host_modules/console_mirror.py
+++ b/host_modules/console_mirror.py
@@ -103,7 +103,7 @@ def printable_escape(payload: bytes) -> str:
         if byte < 0x20 or 0x7F <= byte <= 0x9F:
             return r"\x{:02x}".format(byte)
         return None
-    
+
     def _utf8_width(byte: int) -> int:
         if 0xC2 <= byte <= 0xDF:
             return 2
@@ -615,7 +615,13 @@ class RecordingArchiver:
 
 
 class MirrorManager:
-    """Thread-safe per-line mirror state machine."""
+    """Coordinate one console line's mirror session and runtime state.
+
+    Construction resets the line to ``idle`` in STATE_DB. Use :meth:`start`,
+    :meth:`submit`, :meth:`update_timeout`, :meth:`status`, and :meth:`stop` to
+    manage a session, then call :meth:`shutdown` when the proxy exits. Public
+    methods are thread-safe.
+    """
 
     def __init__(
         self,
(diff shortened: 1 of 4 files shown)


host_modules/console_mirror.py
```diff
@@ -0,0 +1,1336 @@
...
+            self.on_fatal(self, error)
+
+    def close(self) -> None:
+        """Stop accepting records and wait boundedly for queued writes to finish."""
+        with self._lock:
+            self._accepting = False
+            self._shutdown_deadline = time.monotonic() + self.shutdown_timeout
+            self._closing.set()
+        if threading.current_thread() is not self._thread:
+            self._thread.join(self.shutdown_timeout + 0.5)
```
lizhijianrd, line 466: `join()` can return when its timeout expires while the writer thread is still running. `MirrorManager.stop()` may then archive and delete files that are still being written, producing an incomplete ZIP. Could `close()` verify that the thread exited, using `is_alive()` or a completion event, and skip archiving while shutdown is incomplete so the source logs are preserved?

lizhijianrd: We may need a unittest for this scenario too.

lizhijianrd (approved):
The semgrep is a false positive. By design the folder permission should be 0700.

host_modules/console_mirror.py
```diff
@@ -0,0 +1,1343 @@
...
+def _ensure_secure_directory(path: str) -> None:
+    """Ensure that the given path exists and is a secure directory (700 permissions)."""
+    os.makedirs(path, mode=0o700, exist_ok=True)
+    info = os.lstat(path)
+    if not stat.S_ISDIR(info.st_mode) or stat.S_ISLNK(info.st_mode):
+        raise MirrorError(
+            "unsafe_recording_path",
+            f"Recording directory is not a real directory: {path}",
+        )
+    os.chmod(path, 0o700)
```
lizhijianrd:
```
os.chmod(path, 0o700)  # nosemgrep
```

PR #395, opened 2026-06-18
Add an inline platform check for ExecCondition of gnoi-shutdown.service

BYGX-wcr:
Add an inline platform check for ExecCondition of gnoi-shutdown.service so that the gnoi-shutdown.service is skipped on non-smartswitch NPU platforms.

Fix issue: [#27975](https://github.com/sonic-net/sonic-buildimage/issues/27975)

PR #415, opened 2026-07-28
[gnoi-shutdown]: Use Wants= for gnmi.service so ExecCondition runs when gnmi is masked

### LINKED ISSUE #28387: Regression: gnoi-shutdown condition never runs when required delayed gnmi is masked, blocking SYSTEM_READY and sFlow for 180s
### Is it platform specific

mellanox

### Importance or Severity

Critical

### Previous Working Version

202511-based images were reported unaffected. The regression is present in 202605-based images containing https://github.com/sonic-net/sonic-host-services/pull/343 and the later inline `ExecCondition` fix from `sonic-host-services#395`.

### Steps to Reproduce

1. Boot a SONiC platform whose `gnmi` feature is delayed until `PortInitDone`.
2. Wait for `PortInitDone` and for `featured` to start delayed services, including gnmi.
3. Check:

   ```bash
   systemctl status gnoi-shutdown gnmi --no-pager
   systemctl show gnoi-shutdown \
       -p ActiveState -p SubState -p Result \
       -p ConditionResult -p ConditionTimestampMonotonic
   sysreadyshow --detail
   redis-cli -n 6 HGET 'SYSTEM_READY|SYSTEM_STATE' Status
   redis-cli -n 6 HGETALL 'ALL_SERVICE_STATUS|gnoi-shutdown'
   ```

4. Enable and configure sFlow with a collector and interface sampling session.
5. Generate traffic that should produce sFlow samples or drop-monitor reports.

The global readiness failure has been reproduced on:

- SmartSwitch: `x86_64-nvidia_sn4280-r0`
- Non-SmartSwitch: `x86_64-mlnx_msn4700-r0`
- Non-SmartSwitch: `x86_64-mlnx_msn2700a1-r0`

The SN4700 and SN2700 cases reproduce immediately after a fresh 202605 installation, so this is not limited to repeated config reloads or SmartSwitch platforms.

In all three cases:

- gnmi eventually becomes active and healthy.
- `gnoi-shutdown.service` remains `Down/Inactive`.
- No gnoi shutdown daemon process is present.
- `SYSTEM_READY|SYSTEM_STATE` remains absent.

On a live SN2700 reproduction, the enabled gnoi unit reported:

```text
ActiveState=inactive
SubState=dead
Result=success
ConditionResult=no
ConditionTimestampMonotonic=0
```

The zero condition timestamp proves that the inline platform `ExecCondition` was never executed. `ConditionResult=no` alone is only the default value for an unevaluated condition.

### Impact of this regression

Healthd observes `gnoi-shutdown` as the only down service and never publishes `SYSTEM_READY|SYSTEM_STATE Status=UP`.

hsflowd waits for that key to become `UP`, or waits for its 180-second fallback. Consequently, sFlow sampling and drop-monitor reporting can be unavailable for approximately three minutes after every hsflowd start even though sFlow configuration has already reached APPL_DB.

The same missing global readiness state can affect any other component that gates initialization on `SYSTEM_READY`.

### Suspected root cause

`sonic-host-services#343` removed `wait-for-sonic-core.sh` and changed `gnoi-shutdown.service` to use:

```ini
Requires=database.service gnmi.service
Wants=network-online.target pmon.service
After=network-online.target database.service gnmi.service pmon.service
```

gnmi is initially masked/stopped because it is a delayed feature. Systemd resolves the hard `Requires=gnmi.service` dependency before executing gnoi's `ExecCondition`. Therefore:

1. The gnoi start job cannot reach its platform condition while gnmi is masked.
2. The condition remains unevaluated, with `ConditionTimestampMonotonic=0`.
3. Featured later unmasks and starts gnmi after `PortInitDone`.
4. Starting gnmi later does not retry or pull in the inactive gnoi service.
5. Healthd continues to report gnoi as `Down/Inactive` and blocks `SYSTEM_READY`.

The inline condition from `sonic-host-services#395` is therefore insufficient. It can skip gnoi on a non-SmartSwitch only if the service start job reaches the condition in the first place.

The running healthd implementation intentionally requires both `ConditionResult=no` and a nonzero `ConditionTimestampMonotonic` before treating an inactive conditionally skipped service as healthy. This is correct: treating an unevaluated default `no` as a successful skip could hide genuinely unstarted services.

### Expected behavior

- On non-SmartSwitch platforms, gnoi's platform condition should execute and the inapplicable service should not block global readiness.
- On SmartSwitch NPU platforms, gnoi-shutdown should start after delayed gnmi becomes available.
- Once all applicable services are healthy, healthd should publish `SYSTEM_READY|SYSTEM_STATE Status=UP`.
- hsflowd should pass its readiness gate without waiting for the 180-second fallback.

Possible fix directions:

1. Remove or weaken the hard `Requires=gnmi.service` relationship so `ExecCondition` can run while gnmi is masked.
2. Enable/install gnoi-shutdown only on applicable SmartSwitch NPU platforms.
3. On SmartSwitch platforms, arrange for starting delayed gnmi to pull in or retry gnoi-shutdown in the correct order.

### Relevant log output

```shell
# Representative fresh-boot ordering
system-health started
gnmi marked delayed for port initialization
multi-user.target reached
PortInitDone
featured released delayed services
featured unmasked/enabled/started gnmi
gnmi active and healthy

# State remains unchanged many minutes later
SYSTEM_READY|SYSTEM_STATE: absent

ALL_SERVICE_STATUS|gnoi-shutdown:
  service_status: Down
  app_ready_status: Down
  fail_reason: Inactive

ALL_SERVICE_STATUS|gnmi:
  service_status: OK
  app_ready_status: OK

# All other monitored services: OK
```

```shell
Affected SONiC branch: 202605
Affected ASIC vendor: mellanox
Affected platform classes: SmartSwitch and non-SmartSwitch
```

### Attach files (if any)

No files attached. This issue intentionally contains only sanitized status and timestamp evidence.

vvolam:
#### Why I did it
Fixes sonic-net/sonic-buildimage#28387

`gnoi-shutdown.service` declared `Requires=gnmi.service`, but `gnmi` is a delayed feature that is masked at boot until `featured` starts it after `PortInitDone`. With a hard `Requires=` on a masked unit, systemd refuses the gnoi-shutdown start job and never evaluates the unit's `ExecCondition`, leaving `ConditionTimestampMonotonic=0` and `Result=success`.

`healthd` (system-health `sysmonitor`) cannot distinguish this from a genuinely un-started service, so it reports `gnoi-shutdown` as `Inactive` and never publishes `SYSTEM_READY|SYSTEM_STATE Status=UP`. `hsflowd` then waits for that key (or its 180s fallback), so sFlow / drop-monitor is unavailable for ~3 minutes after every start. The inline `ExecCondition` added in #395 cannot help, because the start job never reaches the condition.

Reproduced on non-SmartSwitch `x86_64-mlnx_msn4700-r0` / `msn2700a1-r0` and SmartSwitch `x86_64-nvidia_sn4280-r0`.

#### How I did it
Move `gnmi.service` from `Requires=` to `Wants=` in `data/debian/sonic-host-services-data.gnoi-shutdown.service`, keeping `After=...gnmi.service...` for ordering.

- **Non-SmartSwitch:** a masked `gnmi` is now tolerated, so the start job proceeds and the `ExecCondition` runs and exits non-zero -> `Result=exec-condition`, `ConditionResult=no`, nonzero `ConditionTimestampMonotonic`. healthd treats it as a non-blocking platform skip.
- **SmartSwitch NPU:** `ExecCondition` exits 0 and the daemon starts and listens on STATE/CONFIG_DB. It only needs `gnmi` at DPU-shutdown time (`docker exec gnmi gnoi_client ...`), which is long after `gnmi` is up, so starting early is safe and avoids missing early events.
- `After=` still orders gnoi-shutdown after gnmi whenever gnmi is actually pulled into the transaction (i.e. when it is not masked).

#### How to verify it
On a non-SmartSwitch Mellanox platform (e.g. msn2700) with a fresh 202605 image:
1. Boot and wait for `featured` to start the delayed `gnmi`.
2. `systemctl show gnoi-shutdown -p ActiveState -p SubState -p Result -p ConditionResult -p ConditionTimestampMonotonic`
   - Expect `Result=exec-condition`, `ConditionResult=no`, and a **nonzero** `ConditionTimestampMonotonic` (previously `Result=success`, ts `0`).
3. `sysreadyshow --detail` and `redis-cli -n 6 HGET 'SYSTEM_READY|SYSTEM_STATE' Status` -> `UP`.
4. `redis-cli -n 6 HGETALL 'ALL_SERVICE_STATUS|gnoi-shutdown'` -> no longer blocking.
5. Confirm hsflowd no longer waits the 180s fallback before sFlow samples flow.

On a SmartSwitch NPU (sn4280): confirm `gnoi-shutdown` is active after boot and a DPU `admin_status=down` still triggers the gNOI HALT sequence.

#### Verification results (hardware, 202605, mellanox)
Reproduced the regression and validated the fix on two SKUs. The delayed `gnmi` feature was masked to recreate the exact boot-time condition, then only `Requires=`->`Wants=` was changed:

SmartSwitch NPU `x86_64-nvidia_sn4280-r0` (gnoi-shutdown is applicable):
- Before (`Requires=gnmi.service`, gnmi masked): `systemctl start gnoi-shutdown` -> `Failed to start gnoi-shutdown.service: Unit gnmi.service is masked.`; unit stays `inactive/dead`.
- After (`Wants=gnmi.service`, gnmi still masked): start succeeds -> `active/running`; the platform `ExecCondition` is evaluated (nonzero `ConditionTimestampMonotonic`).

Non-SmartSwitch `Mellanox-SN2700` (gnoi-shutdown is inapplicable):
- Before (`Requires=gnmi.service`, gnmi masked): `Failed to start ... Unit gnmi.service is masked.`; healthd `ALL_SERVICE_STATUS|gnoi-shutdown = Down` and `SYSTEM_READY|SYSTEM_STATE = DOWN`.
- After (`Wants=gnmi.service`, gnmi still masked): start succeeds and the `ExecCondition` now runs and cleanly skips the service -> `Result=exec-condition`, `ConditionResult=no`; healthd `ALL_SERVICE_STATUS|gnoi-shutdown = OK` (non-blocking via `NON_BLOCKING_INACTIVE_REASONS={"exec-condition"}`).

In both cases the only change was `Requires=`->`Wants=`, which lets the start job proceed while `gnmi` is masked so the platform `ExecCondition` actually runs.

- [x] 202605

Reason: boot-time readiness regression introduced in 202605 (via #343 / #395); it blocks `SYSTEM_READY` and delays sFlow on Mellanox SmartSwitch and non-SmartSwitch platforms.

[gnoi-shutdown] Use Wants= instead of Requires= for gnmi.service so the platform ExecCondition runs when gnmi is a masked delayed feature, unblocking SYSTEM_READY.

### DIFF
--- a/data/debian/sonic-host-services-data.gnoi-shutdown.service
+++ b/data/debian/sonic-host-services-data.gnoi-shutdown.service
@@ -1,7 +1,7 @@
 [Unit]
 Description=gNOI based DPU Graceful Shutdown Daemon
-Requires=database.service gnmi.service
-Wants=network-online.target pmon.service
+Requires=database.service
+Wants=network-online.target gnmi.service pmon.service
 After=network-online.target database.service gnmi.service pmon.service
 
 [Service]


dgsudharsan (approved):

stephenxs (approved):

vaibhavhd:
@donghaolicd  please check

markx-arista:
Can we backport this to 202605? This is blocking backport of https://github.com/sonic-net/sonic-host-services/pull/418

vaibhavhd:
I think this fix addresses the sflow/drop monitor issue for now, but the issue here highlights the weakness - anything that waits on `SYSTEM_READY|SYSTEM_STATE` will get impacted due to some/unrelated service is hung/masked/etc.  This design forces every service author to get their dependency declarations exactly right (like this PR's fix), or the entire system pays with a 3-minute readiness penalty.

So I think while we are fixing the issue here, the real gap is still open and can hurt some other service later in the future.

PR #398, opened 2026-06-23
Changes to accept SSH/SNMP/Telemetry from GIL IP for multi-asic platforms

abdosi:
## Description
For multi-asic platforms, add `FORWARD` chain iptables rules to allow SSH/SNMP/Telemetry connections from the GIL (Global In-band Link) IP (namespace management IP) in non-default namespaces.

## Changes
- Add `FORWARD` ACCEPT rules sourced from `namespace_mgmt_ip`/`namespace_mgmt_ipv6` for each ACL service with `multi_asic_ns_to_host_fwd=True` in non-default namespaces
- Change ACL rule chain from `INPUT` to `FORWARD` for non-default namespace services with `multi_asic_ns_to_host_fwd=True`
- Add `FORWARD` DROP rules in non-default namespaces when control plane ACLs are configured

## Motivation
On multi-asic chassis platforms, connections from the GIL IP are forwarded through the namespace network stack rather than directly received as INPUT, so FORWARD chain rules are needed to allow and restrict these connections properly.

PR #422, opened 2026-08-06
Revert "Changes to accept SSH/SNMP/Telemetry from GIL IP for multi-asic platforms (#398)"

yijingyan2:
This reverts commit 65818913f9b391bbd299fd07f63df55fc0e12489.


abdosi:
should not be needed to revert as https://github.com/sonic-net/sonic-mgmt/pull/26859/changes skips the test for now

PR #423, opened 2026-08-07
gnoi-shutdown: use TLS for DPU calls

hdwhdw:
## Why

Certificate-free DPU telemetry servers use ephemeral self-signed TLS. The shutdown daemon must use TLS rather than plaintext and remain robust when the configured legacy port is unavailable.

Tracks sonic-net/sonic-buildimage#28540.

Microsoft ADO (number only): 39179388

## What

- Change DPU gNOI calls from `-notls` to `-insecure`.
- Probe configured, `8080`, and native `50052` ports with bounded, side-effect-free `System.Time`, deduplicated in preference order.
- Document that two unreachable ports can add up to 20 seconds to graceful shutdown.
- Send `Reboot` exactly once on the selected port and poll `RebootStatus` there.
- Clear the halt flag and report failure when no port responds.

## Validation

- The focused suite previously passed 37 tests using import-only stubs because SONiC Python packages are unavailable on the host.
- The added all-ports-fail test verifies `False` return and halt-flag cleanup.
- Python compilation, review, and `git diff --check` pass for head `3677e1e`.
- Integrated physical validation passed through the related utility/DPUProxy TLS path; this daemon was not exercised there.

## Remaining Validation

- Required PR CI is rerunning for head `3677e1e`.

### DIFF
--- a/scripts/gnoi_shutdown_daemon.py
+++ b/scripts/gnoi_shutdown_daemon.py
@@ -246,7 +246,7 @@ class GnoiRebootHandler:
         reboot_cmd = [
             "docker", "exec", "gnmi", "gnoi_client",
             f"-target={dpu_ip}:{port}",
-            "-logtostderr", "-notls",
+            "-logtostderr", "-insecure",
             "-module", "System",
             "-rpc", "Reboot",
             "-jsonin", json.dumps({"method": REBOOT_METHOD_HALT, "message": "Triggered by SmartSwitch graceful shutdown"})
@@ -263,7 +263,7 @@ class GnoiRebootHandler:
         status_cmd = [
             "docker", "exec", "gnmi", "gnoi_client",
             f"-target={dpu_ip}:{port}",
-            "-logtostderr", "-notls",
+            "-logtostderr", "-insecure",
             "-module", "System",
             "-rpc", "RebootStatus"
         ]
--- a/scripts/gnoi_shutdown_daemon.py
+++ b/scripts/gnoi_shutdown_daemon.py
@@ -25,7 +25,7 @@ HALT_IN_PROGRESS_POLL_INTERVAL_SEC = 5  # delay between halt_in_progress checks
 STATUS_RPC_TIMEOUT_SEC = 10  # per RebootStatus RPC timeout
 REBOOT_METHOD_HALT = 3  # gNOI System.Reboot method: HALT
 CONFIG_DB_INDEX = 4
-DEFAULT_GNMI_PORT = "8080"  # Default GNMI port for DPU
+COMMON_GNMI_PORTS = ("8080", "50052")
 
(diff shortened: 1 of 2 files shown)


hdwhdw:
All 37 focused Host Services tests passed, including the 8080/50052 port-selection behavior.

Host Services `Reboot` was not invoked during the physical run. Supporting hardware evidence is limited to the packaged utility observing port 8080 closed and 50052 open, selecting 50052, and completing gNOI halt before a hardware reset.

scripts/gnoi_shutdown_daemon.py
```diff
@@ -241,12 +252,28 @@ def _wait_for_gnoi_halt_in_progress(self, dpu_name: str, state_db=None) -> bool:
...
+        """Return the first DPU port that responds to side-effect-free System.Time."""
+        for port in ports:
+            probe_cmd = [
+                "docker", "exec", "gnmi", "gnoi_client",
+                f"-target={dpu_ip}:{port}",
+                "-logtostderr", "-insecure",
+                "-module", "System",
+                "-rpc", "Time",
+            ]
+            rc, out, err = execute_command(probe_cmd, timeout_sec=STATUS_RPC_TIMEOUT_SEC)
```
qiluo-msft, line 269: each port probe blocks for `STATUS_RPC_TIMEOUT_SEC=10s`. worst case (both ports unreachable) adds 20s to shutdown latency before `_find_working_port` returns None. fine for graceful shutdown but worth documenting in the function docstring so operators know what to expect when a DPU's gNMI is slow to come up.

scripts/gnoi_shutdown_daemon.py
```diff
@@ -241,12 +252,28 @@ def _wait_for_gnoi_halt_in_progress(self, dpu_name: str, state_db=None) -> bool:
...
+            if rc == 0:
+                return port
+            logger.log_warning(f"{dpu_name}: gNMI probe failed (rc={rc}, target={dpu_ip}:{port}): {err}")
+        return None
```
qiluo-msft, line 273: nit: no unit test for the all-ports-fail path (both probes return non-zero → `None` → `_handle_transition` clears halt flag and returns False). `test_find_working_port_falls_back_to_native_port` covers partial failure; a complementary `test_find_working_port_all_ports_fail` would pin the cleanup path.

scripts/gnoi_shutdown_daemon.py
hdwhdw: done

scripts/gnoi_shutdown_daemon.py
hdwhdw: done

hdwhdw:
@qiluo-msft All review threads are resolved. All required CI checks pass. Please review again.

qiluo-msft (approved):

